PR #4, opened 2017-10-16
General doc fixes

BradErz:


.gitignore
```diff
@@ -1,8 +1,8 @@
 /.generate_exes
 /.generate_files
-/tarmak_linux_amd64
-/tarmak_darwin_amd64
-/wing_linux_amd64
-/dist/
-/bin/
-.DS_Store
+tarmak_linux_amd64
```
simonswine: Can you explain, why this is necessary for you. It works fine for me on linux. The `/` forces it to be in the root of the git repo

simonswine:
I merged d69d46e

PR #14, opened 2017-10-24
Tarmak is now compiled against k8s.io release-1.8 branches

wallrj:
Pins k8s.io dependencies to release-1.8 branches which are hopefully more stable, better tested and better integrated than picking a random commit from the master branch of each project.

Fixes: https://gitlab.jetstack.net/tarmak/tarmak/issues/138

I've got this to a point where:

* wing compiles
* I can run wing server inside minikube.
* I can connect to the server using curl

I'd like to write some sort of test to see that it still interacts correctly with wing agent but I think @kragniz may be doing that in another branch. 

Some things I don't understand:

* Why did the API server use sharedinformer factory before? Is it something to do with commented out admission control code?
* If I try to start the wing api server on my machine it prompts for KUBERNETES_* environment variables and tries to read /var/run/secrets/kubernetes.io/serviceaccount/token. This is a standalone API server so why does it need that configuration?

```release-note
Tarmak is now compiled against k8s.io release-1.8 branches.
```


pkg/wing/apiserver/apiserver.go
```diff
@@ -56,28 +56,28 @@ type WingServer struct {
...
 func (c *Config) SkipComplete() completedConfig {
-	return completedConfig{c}
+	completedCfg := c.GenericConfig.Complete()
+	return completedConfig{Config: c, completedConfig: &completedCfg}
 }
 
 // New returns a new instance of WingServer from the given config.
 func (c completedConfig) New() (*WingServer, error) {
-	genericServer, err := c.Config.GenericConfig.SkipComplete().New("wing", genericapiserver.EmptyDelegate) // completion is done in Complete, no need for a second time
+	genericServer, err := c.completedConfig.New("tarmak", genericapiserver.EmptyDelegate) // completion is done in Complete, no need for a second time
```
munnerz: Why rename `wing`->`tarmak`?

pkg/wing/server/server.go
```diff
@@ -91,28 +88,10 @@ func (o WingServerOptions) Config() (*apiserver.Config, error) {
...
 	}
-	informerFactory := informers.NewSharedInformerFactory(client, serverConfig.LoopbackClientConfig.Timeout)
-	admissionInitializer, err := winginitializer.New(informerFactory)
-	if err != nil {
-		return nil, err
-	}
-	if err := o.Admission.ApplyTo(serverConfig, admissionInitializer); err != nil {
-		return nil, err
-	}
-
```
munnerz: Why have you removed this? This is required for admission controllers in wing. Pinning to release branches shouldn't require deleting much code, just changing some types around.

pkg/wing/server/server.go
```diff
@@ -131,11 +110,5 @@ func (o WingServerOptions) RunWingServer(stopCh <-chan struct{}) error {
 		return err
 	}
 
-	/*server.GenericAPIServer.AddPostStartHook("start-sample-server-informers", func(context genericapiserver.PostStartHookContext) error {
-		config.GenericConfig.SharedInformerFactory.Start(context.StopCh)
-		return nil
-	})
-	*/
```
munnerz: Didn't realise this was commented out, and feel like it shouldn't be commented out... but that's something for another PR. Can you just leave this comment block as-is in this PR? No reason to delete here.

hack/update-client-gen.sh
```diff
@@ -6,35 +6,26 @@ set -o errexit
...
+              github.com/jetstack/tarmak/pkg/apis \
+              wing:v1alpha1 \
+              --output-base "${GOPATH}/src/" \
+              --go-header-file ${SCRIPT_ROOT}/hack/boilerplate/boilerplate.go.txt
+
+# We have to run conversion-gen separately so we can set the --extra-peer-dirs
+# flag to not include k8s.io/kubernetes packages (https://github.com/kubernetes/kubernetes/issues/54301)
+
+${CODEGEN_PKG}/generate-internal-groups.sh "conversion" \
+              github.com/jetstack/tarmak/pkg/wing \
```
munnerz: In keeping with other projects, can we move the clients into the `client` subdirectory? (e.g. `github.com/jetstack/tarmak/pkg/wing/client/{clientset,informers,listers}`)

hack/update-client-gen.sh
```diff
@@ -6,35 +6,26 @@ set -o errexit
...
-          --go-header-file "${REPO_ROOT}/hack/boilerplate/boilerplate.go.txt" \
-          --input-dirs="github.com/jetstack/tarmak/pkg/apis/wing" \
-          --input-dirs="github.com/jetstack/tarmak/pkg/apis/wing/v1alpha1" \
-          --internal-clientset-package "github.com/jetstack/tarmak/pkg/wing/clients/internalclientset" \
-          --versioned-clientset-package "github.com/jetstack/tarmak/pkg/wing/client" \
-          --listers-package "github.com/jetstack/tarmak/pkg/wing/listers" \
-          --output-package "github.com/jetstack/tarmak/pkg/wing/informers"
+${CODEGEN_PKG}/generate-internal-groups.sh \
+              "deepcopy,defaulter,client,informer,lister" \
+              github.com/jetstack/tarmak/pkg/wing \
```
munnerz: Also move into `client` subdir

munnerz:
Regarding running wing locally - you need to disable authorization and authentication to do that. I think this is just a couple of flags when you run the binary 😄 

It's not the default as this is baked into the `k8s.io/apiserver` library. There might be a way to force the default, but not sure. We could do with some extra docs to explain this 😄 

/cc @simonswine

simonswine:
I find this really hard to review. Maybe if we rebase things like that in the future to have a single commit that is containing all dep changes and another that is doing the source code changes.

You are doing quite a few things at once. All the compatibility problems with release-1.8 and wing have been solved as part of this PR:  https://gitlab.jetstack.net/tarmak/tarmak/merge_requests/76/commits
(I was just using glide instead of dep)

I am gonna take your vendoring and apply my commit on top of it and I think then we have a better base for this. I would not change the generation of the clients as part of this PR

pkg/wing/server/server.go
simonswine: When I comment it out there is null pointer expection, never bother as things just work for me

simonswine:
@wallrj thanks for your efforts, I have just merged #15. It would be good to implement the generation you have done in a separate PR.

PR #15, opened 2017-10-24
Update vendored k8s.io packages to target release-1.8/release-5.0 branches

simonswine:

Pins k8s.io dependencies to release-1.8 branches which are hopefully more stable, better tested and better integrated than picking a random commit from the master branch of each project.

Fixes: https://gitlab.jetstack.net/tarmak/tarmak/issues/138
Replaces:  #14 

Compared to #14 this reuses the way how we build the API client and listers/informers. Wing server runs in standalone only mode, we never use delegated auth

**Release note**:

```release-note
Update vendored k8s.io packages to target release-1.8/release-5.0 branches
```


simonswine:
@munnerz I am not too sure how to implement the SkipComplete with this updated code base. Maybe you can give me a hint or at least tell the trade off, why we had it before

pkg/apis/wing/types.go
```diff
@@ -52,7 +52,8 @@ type InstanceStatusManifest struct {
 
 type InstanceList struct {
 	metav1.TypeMeta
-	metav1.ObjectMeta
+	// +optional
+	metav1.ListMeta
```
munnerz: Don't think this is, or should be, optional?

pkg/wing/apiserver/apiserver.go
```diff
@@ -55,29 +55,31 @@ type WingServer struct {
...
-// SkipComplete provides a way to construct a server instance without config completion.
-func (c *Config) SkipComplete() completedConfig {
-	return completedConfig{c}
+	return CompletedConfig{&c}
 }
 
 // New returns a new instance of WingServer from the given config.
 func (c completedConfig) New() (*WingServer, error) {
-	genericServer, err := c.Config.GenericConfig.SkipComplete().New("wing", genericapiserver.EmptyDelegate) // completion is done in Complete, no need for a second time
+	genericServer, err := c.GenericConfig.New("wing", genericapiserver.EmptyDelegate)
```
munnerz: Not using 'SkipComplete' looks good to me, as we already store a reference to the completed config on L68 (https://github.com/jetstack/tarmak/pull/15/commits/40390ed7db67156ea79b4d3a9a0fe30b738d377a#diff-eeee2947c8ccb06990948a4196759b69R68).

In the previous code we threw away the `completedConfig` structure returned from `cfg.GenericConfig.Complete`, which meant we had to re-construct this completed configuration (albeit this time without calling Complete - I assume this was okay because Complete was originally called on a pointer? Not sure as I've not checked the code for `Complete` yet.

pkg/apis/wing/types.go
simonswine: Just copied from the core:
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L3409

pkg/wing/apiserver/apiserver.go
simonswine: Just wanted someone to double check

PR #24, opened 2017-10-31
Allow expected autoscale capacity

djsd123:
Modify InstancePool count function

```release-note
enable minCount and maxCount of stateless instancePools
```

### DIFF
--- a/terraform/amazon/templates/instance_pools/instance_pool_instance.tf.template
+++ b/terraform/amazon/templates/instance_pools/instance_pool_instance.tf.template
@@ -23,7 +23,7 @@ data "template_file" "{{.TFName}}_user_data" {
     tarmak_environment   = "${var.environment}"
 {{ if not .Role.Stateful -}}
     tarmak_hostname      = "{{.Role.Name}}"
-    tarmak_desired_count = "${var.{{.TFName}}_count}"
+    tarmak_desired_count = ""
     tarmak_volume_id     = ""
 {{ else -}}
     tarmak_hostname      = "{{.Role.Name}}-${count.index+1}"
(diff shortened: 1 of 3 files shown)


simonswine:
@djsd123 thanks for your contribution. I have renamed count to min_count where ever applicable, we need to validate these values properly, otherwise it takes a while to see the terraform error.

terraform/amazon/templates/instance_pools/instance_pool_instance.tf.template
```diff
@@ -124,6 +123,15 @@ resource "aws_autoscaling_group" "{{.TFName}}" {
     propagate_at_launch = true
   }
 }
+
+resource "aws_autoscaling_policy" "{{.TFName}}-autoscaling-policy" {
```
simonswine: As far as I see is the scaling policy doing exactly nothing, let me know if that is not true? I would merge this after we remove the scaling policy...

munnerz:
How does adjusting these values play into upgrades and/or automatically
draining of nodes upon the max_count var being reduce to less than the
current number of nodes in the ASG?

terraform/amazon/templates/instance_pools/instance_pool_instance.tf.template
djsd123: @simonswine Yep, spot on. It is indeed superfluous.  Removed as requested.

djsd123:
Hi @munnerz, 

Was that question aimed at me?  Not quite sure I follow.

djsd123:
Ah, good question.

djsd123:
@simonswine Would you prefer I closed this?

simonswine:
Merged as part of #267

PR #28, opened 2017-11-06
Use a master controller to disable source-destination checks for instances

mattbates:
Calico requires [source destination checks](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_NAT_Instance.html#EIP_Disable_SrcDestCheck) to be disabled for IP-in-IP tunnelling.

Currently this check is disabled on each of the worker instances with a `systemd` service, but this requires overly permissive IAM (`ec2:ModifyInstanceAttribute`).

This PR adds a [controller](https://github.com/ottoyiu/k8s-ec2-srcdst) that is run in a pod on a master node that is responsible for disabling the check for all Kubernetes notes. Hence, `ec2:ModifyInstanceAttribute` is only required on master instances.

**Release note**:

```release-note
disable source/destination check on cloud-provider AWS using a controller run on kubernetes masters. No need to authorize worker instances for ec2:ModifyInstanceAttribute anymore.
```


simonswine:
There is no `calico/disable-source-destination.yaml.erb`.

The puppet module changes need to go to https://github.com/jetstack/puppet-module-calico first, then we can change the instance role and pull the module into tarmak

simonswine:
@mattbates please use your own fork the next time

PR #32, opened 2017-11-08
Signal handling for wing #21

### LINKED ISSUE #21: Signal handling for wing
Currently we are not handling signals in wing properly. We should implement at least these signals:

`SIGHUP`: Cause a node to be reconverged
`SIGTERM`: Forward sigterm to puppet subprocess (if exists), make sure wing API server is updated accordingly before exiting

JoshVanL:
Adds signal handling to Wing to handle TERM and HUP

```release-note
Adds signal handling to Wing to handle TERM and HUP, SIGHUP: Cause a node to be reconverged, SIGTERM: Forward sigterm to puppet subprocess (if exists)
```
Fixes #21


simonswine:
@JoshVanL thanks, let me know when you have implemented the API part

simonswine:
@JoshVanL please add a release note about the changes. Thanks!

simonswine:
@JoshVanL why do we need an `action-required` here?

/retest all

simonswine:
Also needs a rebase

Gopkg.lock
```diff
@@ -396,6 +402,12 @@
   revision = "3247c84500bff8d9fb6d579d800f20b3e091582c"
   version = "v1.0.0"
 
+[[projects]]
+  name = "github.com/mholt/archiver"
```
simonswine: @JoshVanL it would be great if we could not use this library. I had it before and it causes some head aches.

We are using and vendoring the docker tools already: https://github.com/jetstack/tarmak/blob/master/vendor/github.com/docker/docker/pkg/archive
 
A note the whole things has been renamed to moby https://godoc.org/github.com/moby/moby/pkg/archive

Otherwise looking good.

PR #37, opened 2017-11-17
Test if subtree puppet modules are pushed to upstream

simonswine:
Avoid divergence between Tarmak and separate puppet modules


simonswine:
@kragniz if you could give it a quick sanity check. Thanks!

simonswine:
ok needs that now, you seem to be on-site @kragniz so:

/lgtm

PR #49, opened 2017-11-22
Raise the timeout for the internal ELB of the API server

simonswine:
```release-note
Raise the master LoadBalancer time out to 3600 seconds
```


simonswine:
This was part of the sonobouy problem @kragniz 

/approve

PR #68, opened 2017-12-06
Upgrade prometheus

simonswine:
This upgrade prometheus to 2.0 and breaks down rules and scrape configs into defined resources

```release-note
Upgrade prometheus monitoring to 2.0 (support for RBAC, customizable scraping + alerting configs)
```


simonswine:
A review would be great (just some sanity check reading, quite big)  @JoshVanL

PR #82, opened 2018-01-19
Generate security groups

### LINKED ISSUE #63: Generate security group rules for AWS from Golang templates
> /kind feature

**What happened**:

Currently all security group rules are written in terraform (AWS specifc code)

We keep the firewall settings in generic Go code and generate the template appropriately:
- SG for AWS
- iptables for on-prem

JoshVanL:
* Generic rules put into rule.go
* Security group terraform generated for stacks
* SG Terraform uploaded to container

Makes rules generic and is generated into terraform when needed

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #63

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/tarmak/stack/rules.go
+++ b/pkg/tarmak/stack/rules.go
@@ -18,9 +18,10 @@ type Service struct {
 }
 
 type Port struct {
-	RangeFrom *uint16
-	RangeTo   *uint16
-	Single    *uint16
+	Identifier *string
+	RangeFrom  *uint16
+	RangeTo    *uint16
+	Single     *uint16
 }
 
 type FirewallRule struct {
@@ -46,6 +47,10 @@ var (
 	blackboxPort   = uint16(9115)
 	wingPort       = uint16(9443)
 	maxPort        = uint16(65535)
+
+	k8sIdentifier       = "k8s"
+	k8sEventsIdentifier = "k8sevents"
+	overlayIdentifier   = "overlay"
 )
 
 func newWingService() Service {
@@ -160,9 +165,9 @@ func newEtcdOverlayService() Service {
 		Name:     "etcd",
 		Protocol: "tcp",
 		Ports: []Port{
-			Port{Single: &k8sPort},
-			Port{Single: &k8sEventsPort},
-			Port{Single: &overlayPort},
+			Port{Single: &k8sPort, Identifier: &k8sIdentifier},
+			Port{Single: &k8sEventsPort, Identifier: &k8sEventsIdentifier},
+			Port{Single: &overlayPort, Identifier: &overlayIdentifier},
 		},
 	}
 }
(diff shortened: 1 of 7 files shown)


pkg/tarmak/stack/rules.go
```diff
@@ -0,0 +1,362 @@
...
+
+import (
+	"fmt"
+)
+
+type Rules struct{}
+
+type FirewallRule struct {
+	Name        string
+	Type        string
```
simonswine: Don't like that name... Maybe `Direction` is better

pkg/tarmak/stack/rules.go
```diff
@@ -0,0 +1,362 @@
...
+	FromPort    int
+	ToPort      int
+	Protocol    string
+	CIDIRBlocks []string
+}
```
simonswine: I think it would be best to add these structs:

```
type Hosts {
  Name string        // to build up a rule name programmatically 
  Role string          // to reference instance pool roles
  CIDR *net.IPnet // to use CIDr
}

type Service struct {
  Name string // service name
  Protocol tcp|udp|icmp|any
  Ports []Port
}

type Port {
 RangeFrom *unit16
 RangeTo *unit16
 Single *unit16
}
```
Then Rules would look like
```
Services []Service
Sources []Hosts
Destination []Hosts
```

And you should be able to build the name dynamically in the templates by looping over the single slices

pkg/tarmak/stack/rules.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+}
+
+func (r *Rules) MasterRules(poolName string) []*FirewallRule {
+	return []*FirewallRule{
+		r.Rule_pool_allow_bgp_from_kube_worker(poolName),
+		r.Rule_pool_allow_ipip_from_kuber_worker(poolName),
+		r.Rule_pool_allow_api_from_kube_woker(poolName),
+	}
+}
```
simonswine: You should be able to replace the bunch of these methods  by some lookup method, that finds rules which have `Role == 'bastion'` in `Sources` instead of `BationRules`

Services for all can be defined as global var

simonswine:
@JoshVanL can you allow maintainers to push to this PR (should be a flag on the right side of this)

I have a few changes to push

JoshVanL:
@simonswine Strange, I did already have that checked. I've invited you to be a collaborator on my fork repo which should fix it. Let me know if it still doesn't work

simonswine:
@JoshVanL sorry I had a mistake in my git push command 😞

simonswine:
@dippynark can you take a look at my comments. Again also the verify boilterplate is failing because we don't have the LICENSE headers in the newly added files

simonswine:
Looking good sofar. I am waiting for a manual sonobuoy run to finish and then I am happy to merge
🎉

simonswine:
Looking good now, everything is green in CI and on a built cluster's prometheus

PR #101, opened 2018-01-31
Fix existing VPC docs

dippynark:
**What this PR does / why we need it**: The docs for deploying tarmak into an existing VPC used an incorrect index for accessing the names of the public subnet resource. This PR fixes that

```release-note
NONE
```

### DIFF
--- a/docs/existing-vpc.rst
+++ b/docs/existing-vpc.rst
@@ -62,9 +62,10 @@ For example, if we have the following infrastructure (notation in terraform)::
 		}
 
 		resource "aws_nat_gateway" "main" {
+		  count         = "${length(aws_subnet.public)}"
 		  depends_on    = ["aws_internet_gateway.main"]
 		  allocation_id = "${aws_eip.nat.id}"
-		  subnet_id     = "${aws_subnet.public.1.id}"
+		  subnet_id     = "${aws_subnet.public.*.id[count.index]}"
 		}
 
 		resource "aws_route_table" "public" {


docs/existing-vpc.rst
```diff
@@ -64,7 +64,7 @@ For example, if we have the following infrastructure (notation in terraform)::
 		resource "aws_nat_gateway" "main" {
 		  depends_on    = ["aws_internet_gateway.main"]
 		  allocation_id = "${aws_eip.nat.id}"
-		  subnet_id     = "${aws_subnet.public.1.id}"
```
simonswine: Not too sure but I guess this will break for length(AZ) > 1. I think 0 would be better or create a NAT per public subnet.

Thoughts @kragniz

simonswine:
Can you both look at this @kragniz @dippynark?

simonswine:
Can you both look at this @kragniz @dippynark?

docs/existing-vpc.rst
kragniz: +1 I'd probably go for a seperate nat for each gateway.

Something like
```terraform
resource "aws_nat_gateway" "main" {
  count         = "${length(aws_subnet.public)}"
  depends_on    = ["aws_internet_gateway.main"]
  allocation_id = "${aws_eip.nat.id}"
  subnet_id     = "${aws_subnet.public.*.id[count.index]}"
}
```

docs/existing-vpc.rst
```diff
@@ -62,10 +62,11 @@ For example, if we have the following infrastructure (notation in terraform)::
 		}
 
 		resource "aws_nat_gateway" "main" {
-		  depends_on    = ["aws_internet_gateway.main"]
-		  allocation_id = "${aws_eip.nat.id}"
-		  subnet_id     = "${aws_subnet.public.1.id}"
-		}
+      count         = "${length(aws_subnet.public)}"
```
charlieegan3: minor one, perhaps best to keep the indentation as tabs?

dippynark: ah yeah I was confused as to why that was happening - my editor converts all tabs. Will fix this

PR #122, opened 2018-03-07
Update consul, vault, and golang dependecies

simonswine:
```release-note
Update consul, vault, and golang dependecies
```


simonswine:
@JoshVanL this should fix the trouble we had with `tarmak kubectl` when compiled with Golang 1.10:

https://github.com/jetstack/vault-helper/pull/8

PR #131, opened 2018-03-16
Ensure tools has been deployed before checking wing status

### LINKED ISSUE #55: Ensure wing is only checked in tools/vault/kubernetes

/kind bug

**What happened**:

When you run tarmak against the network stack only. It tries to check wing's status. 

This should only ever happen once wing is available (stack tool has been deployed)

tarmak cluster apply -S network

JoshVanL:
Fixes bug where running only the network stack would check wing status

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #55

```release-note
NONE
```

### DIFF
--- a/pkg/tarmak/terraform.go
+++ b/pkg/tarmak/terraform.go
@@ -69,10 +69,17 @@ func (t *Tarmak) CmdTerraformApply(args []string, ctx context.Context) error {
 		}
 	}
 
-	// wait for convergance in every mode
-	err := t.Cluster().WaitForConvergance()
-	if err != nil {
-		return err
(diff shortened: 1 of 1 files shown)


simonswine:
I think we should fix that by specifying a stack using `-S` should automatically imply a `-I` flag.

That said we are gonna get rid of the stacks anyhow and replace them using a single terraform stack, so it's not important going forward

simonswine:
No longer necessary due to the stack merge

PR #137, opened 2018-03-21
Put container runtime resources into a separate slice

simonswine:
This ensures we can reserve CPU and memory resources for the system and
the runtime properly in a follow-up PR

```release-note
Setup kubernetes runtime components under cgroup podruntime.slice
```

### DIFF
--- a/puppet/modules/kubernetes/templates/kubelet.service.erb
+++ b/puppet/modules/kubernetes/templates/kubelet.service.erb
@@ -4,6 +4,7 @@ Documentation=https://github.com/GoogleCloudPlatform/kubernetes
 <%= scope.function_template(['kubernetes/_systemd_unit.erb']) %>
 
 [Service]
+Slice=podruntime.slice
 WorkingDirectory=<%= @kubelet_dir %>
 <% if @cloud_provider == 'aws' -%>
 # prevent metadata service access on AWS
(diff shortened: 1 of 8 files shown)


puppet/modules/kubernetes/templates/kubelet.service.erb
```diff
@@ -70,10 +71,8 @@ ExecStart=<%= scope['kubernetes::_dest_dir'] %>/kubelet \
 <% end -%>
 <% if scope.function_versioncmp([scope['kubernetes::version'], '1.6.0']) >= 0 -%>
   --cgroup-driver=<%= @cgroup_driver %> \
-<% if not @osfamily.nil? and @osfamily == 'RedHat' -%>
-  --runtime-cgroups=/systemd/system.slice \
-  --kubelet-cgroups=/systemd/system.slice \
-<% end -%>
+  --runtime-cgroups=/podruntime.slice \
+  --kubelet-cgroups=/system.slice \
```
dippynark: There's seems to be no documentation on what the purpose of `--kubelet-cgroups` and `--runtime-cgroups` are, but from [here](https://kubernetes.io/docs/reference/generated/kubelet/) the kubelet is meant to be run in the cgroups referred to by ` --kubelet-cgroups` but the kubelet service unit assigns the kubelet to the `podruntime` slice. From what I understand, a process can only live in exactly one cgroup in a particular hierarchy, so these two settings seems to contradict eachother

dippynark: There's seems to be no documentation on what the purpose of `--kubelet-cgroups` and `--runtime-cgroups` are, but from [here](https://kubernetes.io/docs/reference/generated/kubelet/) the kubelet is meant to be run in the cgroups referred to by ` --kubelet-cgroups` but the kubelet service unit assigns the kubelet to the `podruntime` slice. From what I understand, a process can only live in exactly one cgroup in a particular hierarchy, so these two settings seems to contradict eachother.

simonswine:
Now everything works fine, could you give it another look @dippynark

PR #140, opened 2018-03-21
Add proposal for a custom vault auth provider

mt-inside:
```release-note
NONE
```


docs/proposals/vault-auth-provider.rst
```diff
@@ -0,0 +1,154 @@
...
+This vault auth plugin will serve exactly one environment (i.e. all kubernetes clusters in the same (provider, region))
+
+On login, our provider will:
+
+- Auth the Instance like the AWS provider's EC2 mode does (can we simply defer to that code?)
+- Match the iam-role attached to the Instance against our Provider's config, using its ARN
+- Make a pki-role from the configured template for that iam-role
+- Make a vault-policy (document) from the configured template for that iam-role, including templating in the name of the pki-role
+
+  - This should have a unique name based on the AWS Instance ID / boot ID / etc
```
munnerz: As an alternative, could something in the `tools` stack respond to new instances being created, and automatically provision the roles etc. ready for the instance? & add some retry/check logic in wing to ensure the correct policy/permissions are associated with the token received?

docs/proposals/vault-auth-provider.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Posibilities
+------------
+
+Due to several Vault limitations, especially in the free version, it's
+not possible to:
+
+- Add a vault-policy (name) to a token after it's been created (by vault ec2 login)
+- Specify a vault-policy expressive enough to re-use one policy for all EC2 Instances (i.e. no variable back-references)
+- Vault Premium has a different policy language called Sentinel, which looks like it could do this
```
munnerz: Depending on the difficulty/reusability of the existing AWS auth provider, could we opt for a simpler implementation that requires Vault Enterprise to use?

I'm not sure whether this can be considered a feature that would be largely of interest to enterprise, and whilst it's nice to have supporting the free version the onus of maintaining an auth plugin for general public consumption should be considered.

docs/proposals/vault-auth-provider.rst
munnerz: Perhaps `wing-controller` (not sure if this exists yet) could watch for `Wing` resources in the apiserver, and auto provision the appropriate node roles in Vault? This would save the dependance on the single tools stack for all clusters (which should make portability between clouds, where they may be just one tools stack in one of the clouds (not sure if this'll be the case, but just in case it is))

docs/proposals/vault-auth-provider.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+Notable items
+=============
+
+Concerns
+--------
+
+-  Huge part of security critical code in our hands
+-  Clean up of roles and templates once they are no longer used
```
munnerz: We will also need to extend the vault auth plugin each time we expand tarmak to support a new cloud, or produce a `tarmak-auth-aws`, `tarmak-auth-gcp`, `tarmak-auth-vsphere`, `tarmak-auth-token` etc plugin.

docs/proposals/vault-auth-provider.rst
simonswine: Vault enterprise's sentinel seems to only have access to the meta data available in the token's metadata. The metadata doesn't contain the IP address or the hostname, so I think even with Vault enterprise we would need to do something to get IP/hostname into the token's metadata

docs/proposals/vault-auth-provider.rst
simonswine: This is true. It think I would rather have a single provider with a different release cycle than tarmak itself (maybe as part of vault-helper, which needs a better name). But IMHO a single plugin binary should support mutliple providers. (See the config endpioint flags that are mounted per instance of the plugin)

docs/proposals/vault-auth-provider.rst
simonswine: Yeah right now wing is not involved, if we scale up a prebacked worker image, before the node becomes healthy and usuable.

Making wing a vault controller, would require a (more secure and HA) setup of wing. Then we would need a tarmak for setting up wing :see_no_evil: 

Giving wing that unneeded

munnerz: I'd have thought a secure deployment of Wing is a goal anyway though?

HA is a consideration, however if Wing were to go unavailable, all existing instances (i.e. ones that have had policies provisioned for them already), would continue to work, and should also be able to renew their tokens just fine.

This means a Wing outage = no *new* instances can be created, but existing ones can reboot etc.

simonswine: Yeah but currently wing is not needed for new instances to come up, it is just an orchestratior for updates and reporting back the instance status.

Not being able to bringing up new instances would make a huge requirement for setting up wing in a HA manner. All other components involved in the bootstrap are HA as of today.

simonswine:
@mt-inside could you please fix this error here: then I think we can merge this

https://jetstack-build-infra.appspot.com/build/jetstack-logs/pr-logs/pull/jetstack_tarmak/140/tarmak-docs-verify/134/

kragniz:
@mt-inside the spellcheck failed, add the extra words here https://github.com/jetstack/tarmak/blob/master/docs/spelling_wordlist.txt

docs/proposals/vault-auth-provider.rst
mt-inside: Sorry, thought I'd written about that: another major concern is that this would be a race. After we know an instance is booting (not necessarily under our control if it's say a recovery), then we have to get the role ready /before/ that instance tries to auth with Vault, as the list of roles in a token is static.

docs/proposals/vault-auth-provider.rst
mt-inside: We'd then also have to get everyone on Vault Enterprise, as I *really* wouldn't wanna support two versions

docs/proposals/vault-auth-provider.rst
mt-inside: Fair point, added to the document

mt-inside:
Thanks @kragniz . A full spell-check is brutal for a tech document!

mt-inside:
rebased as requested

simonswine:
/lgtm 

thanks @mt-inside

PR #146, opened 2018-03-22
Add task to build docs using docker

charlieegan3:

Setting up the python environment for the docs is harder than just building the site in docker - which tarmak users will likely have installed.

This is just another option. It's still possible to run create the docs the usual way.

**Release note**:
```release-note
It is now possible to build the docs in a container with `make docker_html`
```

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -27,9 +27,12 @@ help:
 
 # Make the site's html in docker
 docker_html:
-	mkdir -p _build
-	docker build -t tarmakdocs:latest .
-	docker run -it -v "$$(pwd):/site" -u `stat -c "%u:%g" _build` -w /site tarmakdocs:latest
+	set -x; mkdir -p _build; \
+	docker build -t tarmakdocs:latest .; \
+	container=$$(docker create tarmakdocs:latest); \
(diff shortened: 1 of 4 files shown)


docs/Makefile
```diff
@@ -25,6 +25,12 @@ help:
 %: Makefile .venv
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+# Make the site's html in docker
+docker_html:
+	mkdir -p _build
+	docker build -t tarmakdocs:latest .
+	docker run -it -v "$$(pwd):/site" -u `stat -c "%u:%g" _build` -w /site tarmakdocs:latest
```
simonswine: I would suggest using `docker cp` instead. It is more compatible (mac) than volume mounts. Also doesn't create root owned files.

```
docker create
docker cp docs into
docker start
docker cp build out of it
```

charlieegan3: Ahh yeah, that's nicer. Hopefully 7df1d75dc7d6f3c0659bb4b47dc02f88c580692d is looking better.

PR #148, opened 2018-03-26
Merge terraform stacks

### LINKED ISSUE #110: Implement bastion instance in a Tarmak terraform provider
We need to implement the Tarmak terraform provider as per [this](https://docs.tarmak.io/proposals/terraform-provider.html) proposal. 

The resources created for the provider will consume a Tarmak RPC server to properly coordinate applying cluster configuration. #109 will allow us to test the provider resources outside of the terraform container before moving them into the container

dippynark:
**What this PR does / why we need it**: This PR merges the Terraform stacks which were previously run sequentially by Tarmak. This will allow us to both parallelise Terraform and plan all stacks properly

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #110 

```release-note
Breaking change: We refactored our terraform code base to only consist on a single state file. This will destroy any existing AWS objects (!!!)
```


simonswine:
I have just pushed my rebased to upstream/master state

simonswine:
The next step would fixing the a cluster apply of the hub.

The terraform code generation fails currently:

```
go test -v ./pkg/tarmak
--- FAIL: TestTarmak_Terraform_Generate_Hub (0.43s)
```

Issue #149, opened 2018-03-26
Configure a dynamic hard eviction target in memory pressure situations


simonswine:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:

Currently tarmak's kubelets are configured to do a [hard eviction](https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource) based on kubelets default memory target of 100Mi.

**What you expected to happen**:

We should have a memory hard eviction target configured based on the full node's memory  (e.g. 5% of the available memory

**Anything else we need to know?**:

Puppet has the available node memory in so called facts and they can be used to determine that configuration option.

simonswine:
Fixed in #150

PR #150, opened 2018-03-26
Set kubelet eviction-hard based on total memory

charlieegan3:
The kubelet will default to memory.available<100Mi as the hard eviction
threshold. This makes that figure 5% of the total memory or 100Mi.
Whichever is greater.

Read from the default memory facts to set the kubelet `eviction-hard`
parameter.

Default facts:
https://puppet.com/docs/facter/3.10/core_facts.html#memory

Implements https://github.com/jetstack/tarmak/issues/149

**Release note**:

```release-note
Sets a kubelet hard eviction target to 5% of the node's total memory.
```

### DIFF
--- a/puppet/modules/kubernetes/lib/puppet/functions/five_percent_of_total_ram.rb
+++ b/puppet/modules/kubernetes/lib/puppet/functions/five_percent_of_total_ram.rb
@@ -4,7 +4,7 @@ Puppet::Functions.create_function(:five_percent_of_total_ram) do
   end
 
   def five_percent_of_total_ram(total_bytes)
-    five_percent = (total_bytes * 0.05 / 1_000_000).round
+    five_percent = (total_bytes * 0.05 / 1024**2).round
     default = 100
 
     if five_percent < default
--- a/puppet/modules/kubernetes/lib/puppet/functions/five_percent_of_total_ram.rb
+++ b/puppet/modules/kubernetes/lib/puppet/functions/five_percent_of_total_ram.rb
@@ -1,0 +1,16 @@
+Puppet::Functions.create_function(:five_percent_of_total_ram) do
+  dispatch :five_percent_of_total_ram do
+    param 'Integer', :total_bytes
+  end
+
+  def five_percent_of_total_ram(total_bytes)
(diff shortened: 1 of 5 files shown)


puppet/modules/kubernetes/lib/puppet/functions/five_percent_of_total_ram.rb
```diff
@@ -0,0 +1,16 @@
+Puppet::Functions.create_function(:five_percent_of_total_ram) do
+  dispatch :five_percent_of_total_ram do
+    param 'Integer', :total_bytes
+  end
+
+  def five_percent_of_total_ram(total_bytes)
+    five_percent = (total_bytes * 0.05 / 1_000_000).round
```
simonswine: I think with Mi you would need to divide by `1024^2`

puppet/modules/kubernetes/templates/kubelet.service.erb
```diff
@@ -79,6 +79,7 @@ ExecStart=<%= scope['kubernetes::_dest_dir'] %>/kubelet \
   "--tls-cert-file=<%= @cert_file %>" \
   "--tls-private-key-file=<%= @key_file %>" \
 <% end -%>
+  --eviction-hard=memory.available<<%= @hard_eviction_memory_threshold %>
```
simonswine: I would feel safer with quoting the whole thing `"--eviction[...]%>"`

charlieegan3:
Based on the quick verify failure I need to also have these changes in the module repo: https://github.com/jetstack/puppet-module-kubernetes/pull/22 - so I'm trying that...

charlieegan3:
I have a feeling that the ubuntu acceptance tests are also failing because of the changes to the puppet repo config.

```
I0410 10:55:34.503] Call:  /workspace/./test-infra/jenkins/../scenarios/execute.py --env FIXTURES_YML=.fixtures.yml.local -- make -C puppet/modules/tarmak acceptance-1-9-ubuntu
```

This looks like maybe the submodule isn't there for some reason. Either way, I'm not sure how it could be caused by the changes in this PR...

PR #153, opened 2018-03-28
Kube-apiserver should use endpoint reconciler in 1.9+

simonswine:
```release-note
Enable endpoint reconciler for kubernetes API server 1.9+
```


simonswine:
/approve @dippynark 

Please review

PR #157, opened 2018-04-04
Make sha256 check command working on both Linux and Darwin

milosgajdos:
Gets `make build` actually working on `macOS` platforms o_O

no issue has been opened for this, but it's kinda weird you never tested building the project on `macOS` ¯\_(ツ)_/¯

**Release note**:

```release-note
Makefile uses sha256sum command on both Linux and Darwin platforms to
check the integrity of the build. The problem is macOS does not ship the
sha sum utility with the same name. This commit uses the right utility
name for macOS platforms whilst preserving the original command for
usage on Linux OS.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,14 @@ GOPATH ?= /tmp/go
 
 UNAME_S := $(shell uname -s)
 ifeq ($(UNAME_S),Linux)
+	SHASUM := "sha256sum -c"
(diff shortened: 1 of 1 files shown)


simonswine:
Not having a Mac so not too sure why the other thing didn't work

@dippynark please give it a try and then give a LGTM

PR #185, opened 2018-04-20
Fix kubectl

### LINKED ISSUE #184: kubectl fails if terraform folder does not already exist

/kind bug

**What happened**:
I created a cluster using `tarmak cluster apply` and then on a different machine I ran `tarmak kubectl get nodes`. This returned with the following error:
```
FATA[0004] error getting terraform output: chdir dev-hub/terraform: no such file or directory
```

**What you expected to happen**: To be able to run `kubectl` commands on a different machine to where I run `apply`.

Run `tarmak clusters apply` in a different location to where you run `tarmak kubectl get nodes`

**Environment**:
- Install tools:
- Others:

dippynark:
Fixes #184 

```release-note
Allow `tarmak kubectl` to be run from a different location to where the cluster was created
```

### DIFF
--- a/pkg/tarmak/kubectl/kubectl.go
+++ b/pkg/tarmak/kubectl/kubectl.go
@@ -190,6 +190,11 @@ func (k *Kubectl) ensureWorkingKubeconfig() (interfaces.Tunnel, error) {
 
 	// check if certificates are set
 	if len(authInfo.ClientCertificateData) == 0 || len(authInfo.ClientKeyData) == 0 || len(cluster.CertificateAuthorityData) == 0 {
+
+		if err := k.tarmak.Terraform().Prepare(k.tarmak.Environment().Hub()); err != nil {
+			return nil, fmt.Errorf("failed to prepare terraform: %s", err)
+		}
+
 		if err := k.requestNewAdminCert(cluster, authInfo); err != nil {
 			return nil, err
 		}
--- a/cmd/tarmak/cmd/kubectl.go
+++ b/cmd/tarmak/cmd/kubectl.go
@@ -1,8 +1,6 @@
 // Copyright Jetstack Ltd. See LICENSE for details.
 package cmd
 
-import ()
-
 func init() {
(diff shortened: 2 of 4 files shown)


simonswine:
@dippynark please add a release-note
@JoshVanL can you give that a test

charlieegan3:
I've just tested this - works fine for me.

pkg/tarmak/kubectl/kubectl.go
```diff
@@ -190,6 +190,33 @@ func (k *Kubectl) ensureWorkingKubeconfig() (interfaces.Tunnel, error) {
...
+			},
+			nil,
+			nil,
+			nil,
+		); err != nil {
+			return nil, err
+		}
+		// write ssh_config
+		k.tarmak.SSH().WriteConfig()
+
```
simonswine: This is a lot duplicated code from https://github.com/jetstack/tarmak/blob/master/pkg/terraform/terraform.go#L108

This code should be happening in a single method in terraform.go and we just export and call a single method `Prepare()`

pkg/tarmak/kubectl/kubectl.go
```diff
@@ -190,6 +190,14 @@ func (k *Kubectl) ensureWorkingKubeconfig() (interfaces.Tunnel, error) {
 
 	// check if certificates are set
 	if len(authInfo.ClientCertificateData) == 0 || len(authInfo.ClientKeyData) == 0 || len(cluster.CertificateAuthorityData) == 0 {
+
+		if err := k.tarmak.Terraform().Prepare(k.tarmak.Environment().Hub()); err != nil {
+			return nil, fmt.Errorf("failed to prepare terraform: %s", err)
+		}
+
+		// write ssh_config
+		k.tarmak.SSH().WriteConfig()
```
simonswine: Why are we doing that here again, it is already in https://github.com/jetstack/tarmak/blob/master/pkg/tarmak/tarmak.go#L310

If there is a good reason for it please add it to the comments

PR #205, opened 2018-05-10
Install the aws-cli from pip

charlieegan3:
The package from yum install is currently broken and it means we can't bring up new clusters

The aws-cli we're getting from yum is broken at the moment, installing this version with pip seems to work.

Without it we can't bring up vault etc etc

**Release note**:

```release-note
Install awscli in base images using pip to avoid getting broken versions from yum.
```

### DIFF
--- a/packer/amazon/centos-puppet-agent-latest-kernel.json
+++ b/packer/amazon/centos-puppet-agent-latest-kernel.json
@@ -13,21 +13,21 @@
       "secret_key": "{{user `aws_secret_key`}}",
       "region": "{{user `region`}}",
       "source_ami_filter": {
-          "filters": {
-              "owner-alias": "aws-marketplace",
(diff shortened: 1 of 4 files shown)


charlieegan3:
This is WIP as I haven't actually built the images yet.

charlieegan3:
I need to spend more time looking into this but it seems that the kubernetes nodes aren't converging. It seems that there are some issues with wing running puppet - the issue isn't obviously related to anything aws.

simonswine:
This still looks like a problem in test-infra.

/retest

charlieegan3:
Waiting for this fix to go in: https://github.com/jetstack/tarmak/pull/209

charlieegan3:
@simonswine I wasn't able to test thing because of the volume label issue. I was waiting for that to go in before testing this again - just a heads up.

PR #208, opened 2018-05-11
Fix default file path expansion

### LINKED ISSUE #203: Tarmak init creates config file at incorrect path

/kind bug

**What happened**:
I ran tarmak init and it created the config file relative to my current path. e.g.

```
/home/charlieegan3/Code/tarmak/~/.tarmak/tarmak.yaml
```

I expected it to create it in ~/.tarmak/tarmak.yaml

Run tarmak init in a non-HOME directory

MattiasGees:
With the introduction of https://github.com/jetstack/tarmak/pull/194 the default file path didn't work correctly anymore. The filepath was changed to `/~/.tarmak` because of that function. This change fixes that by expanding it before we feed it to the filepath absolute function.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 
fixes #203 

I am not really sure we need to keep the changes of #194. Normally the shell expands an relative directory to a full directory before it gives that to the program. IMO this change was introduced because because one type of shell doesn't do it properly. 

**Release note**:

```release-note
Fix incorrect path for default path of the config file
```

### DIFF
--- a/pkg/tarmak/tarmak.go
+++ b/pkg/tarmak/tarmak.go
@@ -73,15 +73,17 @@ func New(flags *tarmakv1alpha1.Flags) *Tarmak {
 	t.homeDir = homeDir
 
 	// set config directory
+
+	// expand home directory
 	t.configDirectory, err = homedir.Expand(flags.ConfigDirectory)
 	if err != nil {
 		t.log.Fatalf("unable to expand config directory ('%s'): %s", flags.ConfigDirectory, err)
 	}
 
 	// expand relative config path
-	t.configDirectory, err = filepath.Abs(flags.ConfigDirectory)
+	t.configDirectory, err = filepath.Abs(t.configDirectory)
 	if err != nil {
-		t.log.Fatalf("unable to expand relative config directory ('%s'): %s", flags.ConfigDirectory, err)
+		t.log.Fatalf("unable to expand relative config directory ('%s'): %s", t.configDirectory, err)
 	}
 
 	t.log.Level = logrus.DebugLevel


cmd/tarmak/cmd/root.go
```diff
@@ -70,7 +71,10 @@ func stringSliceContains(s []string, e string) bool {
 
 func init() {
 	// set default tarmak config folder
-	tarmakConfigPath := "~/.tarmak"
+	tarmakConfigPath, err := homedir.Expand("~/.tarmak")
+	if err != nil {
+		fmt.Printf("unable to detect home directory: %s", err)
```
dippynark: I feel we should panic here instead of printing that there was a problem - if err is non-nil then tarmakConfigPath won't be set correctly and tarmak will fail more confusingly somewhere later

charlieegan3: `os.Exit(1)` would do I guess.

MattiasGees: I will add `os.Exit(1)` after the Printf so that we have a clear error message and when we shut it down. ok @dippynark @charlieegan3 ?

charlieegan3:
I downloaded this, built the binary and checked that tarmak init and relative paths work.

I think it'd be nice to exit if it fails to work out the directory rather than just printing but this looks good to me.

/lgtm

cmd/tarmak/cmd/root.go
charlieegan3: Sounds good :+1:

MattiasGees: Added it

simonswine:
We already do path expansion in https://github.com/jetstack/tarmak/blob/master/pkg/tarmak/tarmak.go#L76

I should have noticed that bug during my review in #194 :grimacing: 

Will push some commit

simonswine:
Thanks everyone involved

/lgtm

PR #209, opened 2018-05-11
Toggle SELinux volume label based on OS version

charlieegan3:
We've been having an issue using the old label on the latest versions.

Source: https://docs.openshift.com/container-platform/3.6/install_config/persistent_storage/pod_security_context.html

> The volume will be given a type which is accessible by unprivileged containers. This type is usually container_file_t, which treats volumes as container content. Previously, the label specified was svirt_sandbox_file_t. This label is no longer used due to security concerns.

**Release note**:

```release-note
Use the container_file_t volume label on newer os versions.
```

### DIFF
--- a/puppet/modules/kubernetes/manifests/kubelet.pp
+++ b/puppet/modules/kubernetes/manifests/kubelet.pp
@@ -107,11 +107,10 @@ class kubernetes::kubelet(
     $_ca_file = $ca_file
   }
 
-  $os_version = 0 + $facts["operatingsystemrelease"]
-  if $os_version < 7.5 {
-    $seltype = 'svirt_sandbox_file_t'
-  } else {
+  if versioncmp($facts["operatingsystemrelease"], '7.5') >= 0 {
     $seltype = 'container_file_t'
+  } else {
+    $seltype = 'svirt_sandbox_file_t'
   }
 
   file{$kubelet_dir:
--- a/puppet/modules/kubernetes/manifests/kubelet.pp
+++ b/puppet/modules/kubernetes/manifests/kubelet.pp
@@ -107,8 +107,13 @@ class kubernetes::kubelet(
     $_ca_file = $ca_file
   }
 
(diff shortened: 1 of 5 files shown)


charlieegan3:
WIP while I test that `operatingsystemrelease` does what I expect

charlieegan3:
I can run the quick verify command locally and it passes.

```
make -C puppet/modules/tarmak verify
```

For some reason the variable for the os release looks to not be getting set in the test infra run.

puppet/modules/kubernetes/manifests/kubelet.pp
```diff
@@ -107,8 +107,13 @@
     $_ca_file = $ca_file
   }
 
+  $os_version = 0 + $facts["operatingsystemrelease"]
```
simonswine: Can this part behave like the following:

- If fact is existing and os_version < 7.5 use `svirt_sandbox_file_t`
- Else use `container_file_t`

Also for comparing the version, can we please use that: https://github.com/jetstack/tarmak/blob/fee1c17ecb78942474047e0cb7815fc1105269f8/puppet/modules/prometheus/manifests/kube_state_metrics.pp#L19

simonswine:
@charlieegan3 I think to reproduce it locally you might need to set this:

```
export FIXTURES_YML=.fixtures.yml.local
```

puppet/modules/kubernetes/manifests/kubelet.pp
charlieegan3: That's much nicer. I'd got it working with split & more mess but oh well - wish I'd known about this earlier :man_facepalming:

charlieegan3:
I'm able to see the same failures if I run it with the fixtures variable (thanks Christian)

```
FIXTURES_YML=.fixtures.yml.local make -C puppet/modules/tarmak verify
```

Investigating now...

simonswine:
Closed in favour of #211

PR #218, opened 2018-05-12
Use centos with rhel kernel as default image

simonswine:
Upstream kernel is quite risky and can't be maintained in the long term.

```release-note
NONE
```


MattiasGees:
LGTM

PR #80, opened 2017-12-15
WIP: Add PodSecurityPolicies

dippynark:
Fixes #67 

```release-note
Configure a restricted PodSecurityPolicy namespace apart from kube-system and monitoring
 ```

PR #225, opened 2018-05-14
Add PodSecurityPolicies 

### LINKED ISSUE #67: Implement a pod security policy that applies to non cluster addons namespaces
/kind feature

We should use PodSecurityPolicy to limit a users (and controller-managers) ability to create "privileged" (no limited to pod spec flag)  pods out side of namespaces `kube-system` and `monitoring`

System activities on master Nodes typically consume more resources than system activities on worker Nodes so higher values for the --system-reserved flag should be specified.

charlieegan3:
**Copied from https://github.com/jetstack/tarmak/pull/80 so that I can edit the branch.**

Fixes #67

```release-note
Configure a restricted PodSecurityPolicy namespace apart from kube-system and monitoring
```


charlieegan3:
I've tested this on a cluster and it seems to work as expected. I've updated the docs to make sure it's clearer that this is default on regardless of the setting in 1.8.0+ clusters.

PR #228, opened 2018-05-15
Fetch latest selinux package for kubelet config

charlieegan3:
Before configuring the kubelet directory, fetch the latest selinx policy so that the latest labels are available.

```release-note
Running puppet will now always fetch the latest selinux policy package
```


simonswine:
@charlieegan3 seems to listen to me :laughing:

PR #248, opened 2018-05-22
Throw error when we can't find the image ID (AMI) for an instance pool

### LINKED ISSUE #243: When an image does not exist Tarmak does not error

/kind bug

**What happened**: In my `tarmak.yaml`, the instance pool configuration contained `image: centos-puppet-agent-latest-kernel` when the image that was built was called `centos-puppet-agent`. This meant the terraform variable for that image wasn't generated and tarmak errored.

**What you expected to happen**: We should error earlier with a more descriptive error message

Change an instance pool configuration image

MattiasGees:
Before this was implemented we got Terraform errors because the AMI wouldn't be set. This changes makes sure we error earlier with a more decent error message

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #243 

**Release note**:

```release-note
Better error message when an image ID doesn't exist.
```


MattiasGees:
Checking out the failing test.

pkg/tarmak/cluster/cluster.go
```diff
@@ -428,6 +428,8 @@ func (c *Cluster) Variables() map[string]interface{} {
 		ids, ok := imageIDs[image]
 		if ok {
 			output[fmt.Sprintf("%s_ami", instancePool.TFName())] = ids
+		} else {
+			c.log.Fatalf("error getting the image ID of %s", instancePool.TFName())
```
simonswine: I think this needs to happen at another stage before plan/apply commands. We have to be sure that this method is not producing any errors. Furthermore having Fatal calls anywhere in pkg/ should be avoided at any costs

simonswine:
I think it was part of this, but did not survive refactoring: https://github.com/jetstack/tarmak/blob/master/pkg/tarmak/cluster/cluster.go#L151

PR #249, opened 2018-05-22
Ensure templated files are truncated

simonswine:
Currently we don't truncate the files before writing to it. That creates
problems once the file gets shorter than the existing one, as it will
contain extra characters.

```release-note
Ensure templated files are truncated
```

### DIFF
--- a/pkg/terraform/templating.go
+++ b/pkg/terraform/templating.go
@@ -58,7 +58,11 @@ func (t *Terraform) GenerateCode(c interfaces.Cluster) (err error) {
 			"puppet.tar.gz",
 		),
 	)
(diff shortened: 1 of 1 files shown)


charlieegan3:
This looks fine.

I wasn't able to replicate this issue with the terraform module files - as in, making a longer file with invalid HCL at the end didn't cause me errors.

However, if I created a larger puppet.tar.gz (it's usually around 6mb)

```
dd if=/dev/zero of=cluster/terraform/puppet.tar.gz bs=10485760 count=1
```

Then this file would remain 10mb **before this PR**, so it appears to be working as expected.

PR #256, opened 2018-05-24
Implement wing development mode

### LINKED ISSUE #117: Provide a way to build a development tarmak, which by passes prod checks
Goals

- [x] Build tarmak on Mac/Linux using Makefiles so that all latest components are used from this repo
  - [ ] Wing: currently hardcoded
  - [x] Terraform-Provider-Awstag: currently hardcoded

charlieegan3:
Implements the final part of https://github.com/jetstack/tarmak/issues/117

In summary, this allows you to pass a flag when performing terraform actions to enable using a bundled version of the wing binary.

This bundled version is included in the gobindata when building tarmak.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #117

No longer in WIP but I'm open to making changes to the setup here - I think it's unlikely 'finished'.

For example, I wonder if a line in tarmak.yaml might be a better place for this option - rather than a command line flag.

```release-note
Implement a wing development mode that bundles wing binaries.
```


terraform/amazon/templates/puppet_agent_user_data.yaml.template
```diff
@@ -32,6 +33,7 @@ write_files:
       fi ;\
       /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy download --output /opt/wing-$${WING_VERSION}/wing --sha256sums https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt  --signature-armored https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt.asc https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/wing_$${WING_VERSION}_linux_amd64'
     ExecStart=/bin/sh -c 'exec /opt/wing-$${WING_VERSION}/wing agent --manifest-url "s3://${puppet_tar_gz_bucket_path}" --cluster-name "${tarmak_cluster}" --instance-name "$$(curl --silent --retry 5 http://169.254.169.254/latest/meta-data/instance-id || echo "unknown")" --server-url "https://bastion.${tarmak_environment}.${tarmak_dns_root}:9443"'
+{{- end }}
```
simonswine: It would be great if we could use the same `ExecStart=` here and have the binary path defined in a Environment variable.

terraform/amazon/templates/puppet_agent_user_data.yaml.template
```diff
@@ -15,12 +15,13 @@ write_files:
     [Service]
     Environment=AWS_REGION=${region}
     Environment=PATH=/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin:/opt/bin:/root/bin
-    Environment=AIRWORTHY_VERSION=0.2.0
-    Environment=AIRWORTHY_HASH=2d69cfe0b92f86481805c28d0b8ae47a8ffa6bb2373217e7c5215d61fc9efa1d
-    Environment=WING_VERSION=0.3.0
```
simonswine: We need to keep those for the non-dev mode to work

pkg/tarmak/assets/assets.go
```diff
@@ -5,4 +5,4 @@
 
 package assets
 
-//go:generate go-bindata -prefix ../../../ -pkg $GOPACKAGE -o assets_bindata.go ../../../terraform/amazon/modules/... ../../../terraform/amazon/templates/... ../../../puppet/... ../../../packer/...
+//go:generate go-bindata -prefix ../../../ -pkg $GOPACKAGE -o assets_bindata.go ../../../wing_linux_amd64 ../../../terraform/amazon/modules/... ../../../terraform/amazon/templates/... ../../../puppet/... ../../../packer/...
```
simonswine: It would be great if we could have smaller release binaries that do not contain the wing_linux_amd64 built into.

I have had some branch that did not merge, which vendored other binaries using a devmode build tag:

https://github.com/jetstack/tarmak/pull/126/commits/d89de015fef66a33d100f0895801e031a20be72c

I should have remembered that before you started implenting it :grimacing:

terraform/amazon/templates/puppet_agent_user_data.yaml.template
charlieegan3: So upload the a checksum of the dev binary to s3 as well?

What is the format for generating checksums that Airworthy needs?

simonswine: No I mean you just run the same ExecStart. (The ExecStartPre should still be templated through `If/Else`. THat means in dev mode you don't call airworthy.

In general airworthy is downloaded in a specific version (with it specific hardcoded hash). Then after that everything airworthy downloads is GPG signed

PR #263, opened 2018-05-29
Implement an external scrape mode for prometheus

simonswine:
This mode allows a tarmak prometheus deployment, which only contains
scrape_configs for nodes outside of the cluster

**Release note**:

```release-note
Implement an external scrape mode for prometheus
```

### DIFF
--- a/puppet/modules/prometheus/manifests/init.pp
+++ b/puppet/modules/prometheus/manifests/init.pp
@@ -6,15 +6,18 @@ class prometheus(
   Integer[1025,65535] $etcd_k8s_main_port = $::prometheus::params::etcd_k8s_main_port,
   Integer[1025,65535] $etcd_k8s_events_port = $::prometheus::params::etcd_k8s_events_port,
   Integer[1024,65535] $etcd_overlay_port = $::prometheus::params::etcd_overlay_port,
+  Boolean $external_scrape_targets_only = false,
 ) inherits ::prometheus::params
 {
 
   if $role == 'master' {
     include ::prometheus::server
-    include ::prometheus::kube_state_metrics
-    include ::prometheus::node_exporter
-    include ::prometheus::blackbox_exporter
     include ::prometheus::blackbox_exporter_etcd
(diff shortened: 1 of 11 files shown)


puppet/modules/prometheus/manifests/init.pp
```diff
@@ -6,15 +6,19 @@
...
-    include ::prometheus::kube_state_metrics
-    include ::prometheus::node_exporter
-    include ::prometheus::blackbox_exporter
     include ::prometheus::blackbox_exporter_etcd
+    include ::prometheus::node_exporter
+    if ! $external_scrape_targets_only {
+      include ::prometheus::kube_state_metrics
+      include ::prometheus::node_exporter
+      include ::prometheus::blackbox_exporter
+    }
```
dippynark: node_exporter is included in both cases, is this intended?

simonswine: Fixed in latest push

puppet/modules/prometheus/manifests/node_exporter.pp
```diff
@@ -94,11 +68,41 @@
...
-      manifests => [
-        template('prometheus/prometheus-ns.yaml.erb'),
-        template('prometheus/node-exporter-ds.yaml.erb'),
-      ],
+    $external_scrape_targets_only = $::prometheus::external_scrape_targets_only
+    if ! $external_scrape_targets_only {
+      kubernetes::apply{'node-exporter':
+        manifests => [
+          template('prometheus/prometheus-ns.yaml.erb'),
+          template('prometheus/node-exporter-ds.yaml.erb'),
```
dippynark: we include node_exporter when `! $external_scrape_targets_only` is true so I assume this apply should be before the `! $external_scrape_targets_only` check too?

simonswine: No we always include node_exporter now, but only if external_scrape_targets_only is not true we want to have the daemonset node exporters

PR #266, opened 2018-05-29
Ensure environment is passed through for kubectl

### LINKED ISSUE #255: unable to launch the editor "vi"

/kind bug

**What happened**:

When attempting to edit kubernetes resource using `kubectl` tarmak fails with following error:

```
error: unable to launch the editor "vi"
```

I expected the `vi` or `vim` [or whatever other] `$EDITOR` to launch so I can edit the kubernetes resources

```
armak -c ./ kubectl edit clusterroles "psp:privileged"
DEBU[0000] trying to connect to https://127.0.0.1:57588  app=tarmak
DEBU[0000] connected to kubernetes api v1.8.8            app=tarmak
error: unable to launch the editor "vi"
```

I reckon this is happening because the container which is launched when `kubectl` is used does not contain `vi` or `vim` or whatever `$EDITOR` in the container image hence kaboom.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.8.8
- Cloud provider or hardware configuration**: AWS
- Install tools:  🛠
- Others: 👀

simonswine:
Fixes #255

```release-note
NONE
```

### DIFF
--- a/pkg/tarmak/kubectl/kubectl.go
+++ b/pkg/tarmak/kubectl/kubectl.go
@@ -282,9 +282,10 @@ func (k *Kubectl) Kubectl(args []string) error {
 	}
 
 	cmd := exec.Command("kubectl", args...)
-	cmd.Env = []string{
+	cmd.Env = append(
+		os.Environ(),
 		fmt.Sprintf("KUBECONFIG=%s", k.ConfigPath()),
-	}
+	)
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	cmd.Stdin = os.Stdin


pkg/tarmak/kubectl/kubectl.go
```diff
@@ -282,9 +282,10 @@ func (k *Kubectl) Kubectl(args []string) error {
 	}
 
 	cmd := exec.Command("kubectl", args...)
-	cmd.Env = []string{
+	cmd.Env = append(
+		os.Environ(),
 		fmt.Sprintf("KUBECONFIG=%s", k.ConfigPath()),
```
munnerz: Just a thought - if the local user already has KUBECONFIG set to something non-empty, which value will take precedence?

simonswine: From golang stdlib:

```
        // Env specifies the environment of the process.
        // Each entry is of the form "key=value".
        // If Env is nil, the new process uses the current process's
        // environment.
        // If Env contains duplicate environment keys, only the last
        // value in the slice for each duplicate key is used.
        Env []string
```

I think this behaviour changed with golang 1.9+

simonswine:
@munnerz this PR was created after we updated the prow config :cry:

PR #276, opened 2018-06-04
Secure endpoints to only allow certain cidr blocks

MattiasGees:
**What this PR does / why we need it**:
This secures public endpoints (jenkins, bastion, api server) to only allow certain IP ranges.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #198 

**Special notes for your reviewer**: Code is ready, still need to write the docs
**Release note**:

```release-note
Secure public endpoints (bastion, jenkins, apiserver) with cidr ranges
```

TODO:
- [x] Add environment admin cidr + default
- [x] Add default for allowcidr to take the admin cidr
- [x] Test
- [x] Docs

PR #285, opened 2018-06-04
Refactor validate and verify

MattiasGees:
Make sure that validate and verify are present on cluster, environment and provider. This streamlines the use of both validate and verify as central calling point instead of all different methods.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)
I need this for #276 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/tarmak/provider/amazon/amazon.go
+++ b/pkg/tarmak/provider/amazon/amazon.go
@@ -356,6 +356,10 @@ func (a *Amazon) Validate() error {
 
 }
 
+func (a *Amazon) Verify() (result error) {
+	return result
+}
+
 func (a *Amazon) getAvailablityZoneByRegion() (zones []string, err error) {
 	svc, err := a.EC2()
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -356,6 +356,10 @@ func (a *Amazon) Validate() error {
 
 }
 
+func (a *Amazon) Verify() (result error) {
+	return result
```
charlieegan3: Am I right in saying this is the same as returning nil?

charlieegan3: I am https://play.golang.org/p/vJ5A24UnA99

I'm going to stick my neck out and say I prefer `return nil` :fire: 

```
func Verify() (error) {
	return nil
}
```

charlieegan3: (this all looks fine. I'm winding down for the day. Just ignore my wittering...)

MattiasGees: Yeah we use the same methodology on other places as well.

PR #288, opened 2018-06-05
Improve vault tunnel creation

dippynark:
This PR fixes the following:
- when testing connection to the bastion before creating vault tunnels, we use the `-N` flag, which shouldn't be used as we are running a remote command (`/bin/true`)
- when creating vault tunnels, if a problem occurs, there was a chance we could hang indefinitely - this change reduces the chance of a blocking channel so that this hanging doesn't occur

```release-note
Improve Vault tunnel creation preventing indefinite hanging in certain situations
```

### DIFF
--- a/pkg/tarmak/vault/vault.go
+++ b/pkg/tarmak/vault/vault.go
@@ -24,7 +24,7 @@ const (
 	VaultStateUnsealed
 	VaultStateUnintialised
 	VaultStateErr
-	vaultTunnelCreationTimeoutSeconds = 5
+	vaultTunnelCreationTimeoutSeconds = 100
 )
 
 const (
@@ -108,7 +108,7 @@ func (v *Vault) TunnelFromFQDNs(vaultInternalFQDNs []string, vaultCA string) (in
 				return
 			}
 
-			if health.Standby == false && health.Sealed == false {
+			if health.Standby == false && health.Sealed == false && health.Initialized == true {
 				activeNode <- pos
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/tarmak/vault/vault.go
```diff
@@ -109,20 +110,25 @@ func (v *Vault) TunnelFromFQDNs(vaultInternalFQDNs []string, vaultCA string) (in
 
 			if health.Standby == false && health.Sealed == false {
```
simonswine: Maybe we have to check here for health.Initialized

pkg/tarmak/vault/vault.go
```diff
@@ -24,6 +24,7 @@ const (
 	VaultStateUnsealed
 	VaultStateUnintialised
 	VaultStateErr
+	vaultTunnelCreationTimeoutSeconds = 5
```
simonswine: Please test multi/single cluster E2E, I am not too sure if that is a good value

simonswine:
/assign @dippynark 

@dippynark what is your feeling about my comments?

pkg/tarmak/vault/vault.go
dippynark: It worked fine with 5, but have upped it to 10 to be safe

pkg/tarmak/vault/vault.go
dippynark: agreed

simonswine:
Thanks luke for tracking this down

/lgtm
/approve

PR #300, opened 2018-06-12
Change cgroup driver from systemd to cgroupfs

dippynark:
**What this PR does / why we need it**: When using the systemd driver, node allocatable limits are not being set properly by kubelet. Switching driver to cgroupfs fixes this issue.

```release-note
Change cgroup driver from systemd to cgroupfs for docker and kubelet
```

### DIFF
--- a/puppet/modules/kubernetes/manifests/kubelet.pp
+++ b/puppet/modules/kubernetes/manifests/kubelet.pp
@@ -26,7 +26,6 @@ class kubernetes::kubelet(
   $pod_cidr = undef,
   $hostname_override = undef,
   Enum['systemd', 'cgroupfs'] $cgroup_driver =  $::osfamily ? {
-    'RedHat' => 'systemd',
     default  => 'cgroupfs',
   },
   String $cgroup_root = '/',
--- a/puppet/modules/kubernetes/spec/acceptance/single_node_spec.rb
+++ b/puppet/modules/kubernetes/spec/acceptance/single_node_spec.rb
@@ -66,6 +66,9 @@ class{'kubernetes::worker':
       # Ensure docker is setup
       hosts_as('master').each do |host|
         on host, 'yum install -y docker'
+        on host, 'cp -a /usr/lib/systemd/system/docker.service /etc/systemd/system/docker.service'
+        on host, 'sed -i -e \'s/systemd/cgroupfs/g\' /etc/systemd/system/docker.service'
+        on host, 'systemctl daemon-reload'
         on host, 'systemctl start docker.service'
       end
     end
--- a/puppet/modules/site_module/files/20-cgroupfs.conf
+++ b/puppet/modules/site_module/files/20-cgroupfs.conf
@@ -1,0 +1,16 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd-current \
+          --add-runtime docker-runc=/usr/libexec/docker/docker-runc-current \
+          --default-runtime=docker-runc \
+          --exec-opt native.cgroupdriver=cgroupfs \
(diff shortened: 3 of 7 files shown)


puppet/modules/kubernetes/manifests/kubelet.pp
```diff
@@ -26,7 +26,7 @@
   $pod_cidr = undef,
   $hostname_override = undef,
   Enum['systemd', 'cgroupfs'] $cgroup_driver =  $::osfamily ? {
-    'RedHat' => 'systemd',
+    'RedHat' => 'cgroupfs',
```
charlieegan3: can this not just rely on the default if it's now the same?

dippynark: yeah true, will change

puppet/modules/site_module/files/20-cgroupfs.conf
```diff
@@ -0,0 +1,17 @@
...
+          --init-path=/usr/libexec/docker/docker-init-current \
+          --seccomp-profile=/etc/docker/seccomp.json \
+          $OPTIONS \
+          $DOCKER_STORAGE_OPTIONS \
+          $DOCKER_NETWORK_OPTIONS \
+          $ADD_REGISTRY \
+          $BLOCK_REGISTRY \
+          $INSECURE_REGISTRY \
+          $REGISTRIES
+
```
charlieegan3: :policeman: no blank line.

I think the VSCode setting is:

```
"files.insertFinalNewline": true
```

:running_man:

puppet/modules/site_module/files/20-cgroupfs.conf
```diff
@@ -0,0 +1,16 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/dockerd-current \
+          --add-runtime docker-runc=/usr/libexec/docker/docker-runc-current \
+          --default-runtime=docker-runc \
+          --exec-opt native.cgroupdriver=cgroupfs \
+          --userland-proxy-path=/usr/libexec/docker/docker-proxy-current \
+          --init-path=/usr/libexec/docker/docker-init-current \
+          --seccomp-profile=/etc/docker/seccomp.json \
```
simonswine: Have you tried if a second `--exec-opt native.cgroupdriver=cgroupfs` in OPTIONS is able to override the first one @dippynark

simonswine:
I am bit concerned, what the effects are when we update a running cluster with that change.

Can you do some testing @dippynark

puppet/modules/site_module/files/20-cgroupfs.conf
dippynark: yeah, it didn't work

dippynark:
@simonswine I had to restart docker manually but Pods were moved to the new cgroup

simonswine:
Can we make the restart happen through puppet (we should be able to)

/reassign @dippynark

dippynark:
@simonswine let me know when you are happy this works and I will work on fixing the tests

puppet/modules/kubernetes/spec/acceptance/single_node_spec.rb
```diff
@@ -66,6 +66,9 @@
       # Ensure docker is setup
       hosts_as('master').each do |host|
         on host, 'yum install -y docker'
+        on host, 'cp -a /usr/lib/systemd/system/docker.service /etc/systemd/system/docker.service'
+        on host, 'sed -i -e \'s/systemd/cgroupfs/g\' /etc/systemd/system/docker.service'
+        on host, 'systemctl daemon-reload'
```
simonswine: I think you have to do the same thing for `tarmak/spec/acceptance` @dippynark

PR #301, opened 2018-06-12
Fix docker slice drop-in

dippynark:
**What this PR does / why we need it**: The docker slice drop-in is currently creating a directory instead of a file. The quoting is also rendering the content incorrectly. This change fixes this so that docker actually runs under podruntime.slice.

```release-note
Run docker under podruntime.slice
```


dippynark:
test puppet-tarmak-acceptance-centos v1.7

simonswine:
Good spot 

/approve
/lgtm

PR #304, opened 2018-06-13
Fix stateful instance naming

dippynark:
**What this PR does / why we need it**: stateful instances are named incorrectly (including the word k8s) and containing hyphens which was inconsistent with the stateless instances. Additionally, jenkins instances had a special prefix  resulting in `tarmak clusters instances list` not working properly. This PR fixes that

```release-note
NONE
```


simonswine:
Seems sensible to me

/approve

PR #306, opened 2018-06-14
Cluster and instance pool section should within IAM additional chapter

simonswine:
```release-note
NONE
```


MattiasGees:
Correct
/lgtm

PR #307, opened 2018-06-16
Adds set-current cluster subcommand

### LINKED ISSUE #155: Add CLI command to change current cluster
/kind feature

~/.tarmak/tarmak.yaml contains an array of clusters, and a currentCluster key, but there's no way to change the current cluster (used for all operations?) other than to manually edit that hidden file. Some equivalent to kubectl config set-context would be good, some equivalent of kubectx would be even better.

JoshVanL:
Currently the only way to  set the current cluster is to edit the yaml. This adds a new cluster subcommand, set-current, that gives the option to change the current-cluster in the CLI.

fixes #155 

```release-note
Add subcommand to set-current cluster
```

### DIFF
--- a/cmd/tarmak/cmd/cluster_set-current.go
+++ b/cmd/tarmak/cmd/cluster_set-current.go
@@ -1,0 +1,45 @@
+// Copyright Jetstack Ltd. See LICENSE for details.
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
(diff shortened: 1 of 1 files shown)


dippynark:
@JoshVanL nice, we've been needing this

simonswine:
/approve
/lgtm

@JoshVanL thanks for that feature. For future PRs add release notes where ever it makes sense

Issue #280, opened 2018-06-04
Distinguish between Validations and Verifications


simonswine:
Configuration validation/verification currently is done in a pretty error prone and inconsistent way. Moving forward we should clean up this technical debt by distinguishing between validations and verifications. The interfaces for objects in tarmak (Provider/Environment/Cluster/InstancePool/...) should have those methods:

**Validations** `Validate() error`

Validations are offline checks if the configuration is consistent. They should be executed against all objects in a `tarmak.yaml` config

**Verifications** `Verify() error`

Verifications can be online and they are more likely to fail and are quite expensive in terms of time spend . These verifications should only be executed for resources that are associated with the current clusters

Those sub tasks track more specific work around this effort:

EDIT me

Verifications are more

PR #314, opened 2018-06-18
WIP: Flexible instance type

### LINKED ISSUE #310: Change instance type

>
> /kind bug
/kind feature

**What happened**:
At this moment in tarmak.yaml, we can only provide small, medium, large, ...

I would expect we can specify a specific AWS instance like c5.large. This makes tarmak more tweakable for end-user. 

For master I would do a check that we don't launch instances that are too small to run everything (eg t2.nano, t2.micro, ...). A sensible default should also be set. 

Especially for masters we should make this tweakable. People have totally different workloads (eg memory, cpu, gpu, ...)

``
tarmak init``

**Environment**:
- Install tools:
- Others:

JoshVanL:
Although users are already able to chose their own instance types, this validates the supported types and restricts the types available to master

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #310 

```release-note
NONE
```


JoshVanL:
This adds some validation to the instance types and restricts the master instance types.
/cc @MattiasGees  What do you think?

pkg/tarmak/cluster/cluster.go
```diff
@@ -142,27 +146,30 @@ func validateClusterTypes(poolMap map[string][]*clusterv1alpha1.InstancePool, cl
...
 			result = multierror.Append(result, err)
 		}
 	}
-	return result
-}
 
-// Verify cluster
-func (c *Cluster) Verify() (result error) {
-	return c.VerifyInstancePools()
+	if err := c.Environment().Provider().ValidateInstanceTypes(c.InstancePools()); err != nil {
```
MattiasGees: I think this needs to become a verify check and not an validate check. This will contact AWS and is a more expensive call. This is to keep working in line with https://github.com/jetstack/tarmak/issues/280

pkg/tarmak/cluster/cluster.go
```diff
@@ -142,27 +146,30 @@ func validateClusterTypes(poolMap map[string][]*clusterv1alpha1.InstancePool, cl
...
+		result = multierror.Append(result, err)
+	}
+
+	return result.ErrorOrNil()
 }
 
-// Verify instance pools
-func (c *Cluster) VerifyInstancePools() (result error) {
+// Verify cluster
+func (c *Cluster) Verify() error {
```
MattiasGees: Why remove the VerifyInstancePools?

pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -600,6 +621,30 @@ func (a *Amazon) VolumeType(typeIn string) (typeOut string, err error) {
...
+	}
+
 	return typeIn, nil
 }
+
+func (a *Amazon) awsVolumeTypes() []string {
+	return []string{"io1", "gp2", "st1", "sc1"}
+}
+
+func (a *Amazon) awsInstanceTypes() []string {
```
MattiasGees: I think you are missing some newer instance types, is there a way to make this dynamic and check this through an AWS call?

pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -521,20 +516,35 @@ func (a *Amazon) VerifyInstanceTypes(instancePools []interfaces.InstancePool) er
...
-			return err
+			result = multierror.Append(result, err)
+			continue
 		}
 
 		if err := a.verifyInstanceType(instanceType, instance.Zones(), svc); err != nil {
 			result = multierror.Append(result, err)
 		}
+
+		if instance.Name() == "master" {
```
MattiasGees: I think we also should do a check for vault and etcd. I don't think we should run these on burstable (t) types.

pkg/tarmak/cluster/cluster.go
JoshVanL: done

pkg/tarmak/provider/amazon/amazon.go
JoshVanL: I've had a look and I can't find a straight forward way of achieving this. I've updated this list though

pkg/tarmak/provider/amazon/amazon.go
JoshVanL: done

pkg/tarmak/cluster/cluster.go
JoshVanL: done

MattiasGees:
I tested this and this works great. 
I encountered one problem. When we use the default settings we assign to the instances atm we get verification errors due to small instance types :). So I think we should adjust that. 

Currently we set etcd to `small` and vault to `tiny` which translates to an `t2.medium` and `t2.nano`.

For etcd I would switch the default to an `m5.large` but for vault I am not sure we should go that big as we only request SSL certs from time to time and don't do anything heavy on it.

JoshVanL:
I have bumped up etcd instances to use medium (m4.large) as default at init. I think there should be some consensus though on these sizes and what vault should have / be burstable.

MattiasGees:
@JoshVanL I discussed this a bit with @simonswine. We agreed that denying of certain instance types shouldn't happen. 
As for development environments we can see it happen that you run way less and that a burst-able type could work. We agreed instead of giving an error that a warning like `It is not advised to use t2.nano for etcd` would be better. Is that something we can implement in an easy way?

JoshVanL:
@MattiasGees yes that can be done easily

MattiasGees:
@JoshVanL I tried bringing up a cluster with smallest instances for etcd and master, but couldn't see any warnings displayed. So I think there is something wrong with the warning logger.

JoshVanL:
@MattiasGees 
Thanks Mattias for spotting, should be fixed now

MattiasGees:
Works now
/lgtm

PR #321, opened 2018-06-19
Remove Jenkins volume from state module

dippynark:
**What this PR does / why we need it**: A Jenkins volume is created by the state module that isn't actually used by Jenkins (as the Jenkins module creates one). This fixes that.

```release-note
Remove unnecessary Jenkins volume from state module
```


dippynark:
@MattiasGees I've removed that already?

MattiasGees:
correct, for some reason I looked over it
/lgtm

PR #325, opened 2018-06-19
Autoscaling arbitrary worker pools

### LINKED ISSUE #324: Support autoscaling arbitrary work instance pools

/kind bug

**What happened**: Specifying multiple work instance pools does not working - only one gets created and it has to be called worker

**What you expected to happen**: To be able to create multiple work instance pools with any name that will be autoscaled by the cluster autoscaler

**How to reproduce it (as minimally and precisely as possible)**: create a worker instance pool not called worker

dippynark:
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #324 

```release-note
Support creating and autoscaling an arbitrary number of worker instance pools
```

### DIFF
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -83,3 +83,4 @@ autoscaling
 preempted
 millicores
 yaml
+autoscale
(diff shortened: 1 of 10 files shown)


MattiasGees:
@di
I tested this and it works without any problems. Can you just fix the docs, as they still mention that we only support one worker asg.

/assign @di
/unassign

MattiasGees:
@dippynark
I tested this and it works without any problems. Can you just fix the docs, as they still mention that we only support one worker asg.

/assign @dippynark
/unassign

PR #328, opened 2018-06-20
Add kubeconfig command

### LINKED ISSUE #279: Provide a kubeconfig subcommand for clusters

/kind feature

**What happened**:

People/CI jobs tend to use other CLI tools that read the kubernetes config themselves and a kubeconfig could be a good way to make that more approachable.

A kubeconfig subcommand should verify that the cluster is reachable (similar like kubectl subcommand is doing) and output the path to the kubeconfig like that:

```
KUBECONFIG=/full/path/to/kubeconfig
```

It then can be used in a shell session like that:

```
eval "$(tarmak cluster kubeconfig)"
kubectl get nodes
helm list
```

We also should support a public flag that will behave slightly different:

* Error if .kubernetes.apiserver.public is false
* Write config to `kubeconfig-public` and use the ELB public URL instead

MattiasGees:
This adds the kubeconfig command to make sure kubeconfig is present and outputs the kubeconfig path so the user can export it as an environment variable.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #279 

@simonswine I decided to only let this work when the APIserver is public. This because otherwise we will need to keep the SSH tunnel open some way and IMHO this should not be part of this command.
I think something like `tarmak clusters debug kubectl` could be an option that opens a shell with everything set correctly and that tears down the SSH tunnel on exit. What do you think?

**Release note**:

```release-note
Add kubeconfig command
```


simonswine:
I think having a `public-api-endpoint=true` / `public-api-endpoint=false` would provide good value for CI environments where the 10 min timeout of the tunnel doesn't matter so much.

If the flag is not set at all, we should fallback to autodetection. (if public api is available use it).

We should display a warning if the private API endpoint is used, that the tunnel is gonna expire in 10 minutes

/assign @MattiasGees 
/unassign

MattiasGees:
@simonswine this is now implemented for kubeconfig. Do you want me to try implement something like this for the kubectl command as well?

dippynark:
@MattiasGees  This doesn't seem to work for me - in need of this for wargaming:
```
$ tarmak -c . clusters kubeconfig --public-api-endpoint=true
DEBU[0000] trying to connect to https://api.workshop-cluster.tarmak.org  app=tarmak
WARN[0000] error connecting to cluster: Get https://api.workshop-cluster.tarmak.org/version: dial tcp: lookup api.workshop-cluster.tarmak.org: no such host  app=tarmak
DEBU[0000] trying to connect to https://api.workshop-cluster.tarmak.org  app=tarmak
WARN[0000] error connecting to cluster: Get https://api.workshop-cluster.tarmak.org/version: dial tcp: lookup api.workshop-cluster.tarmak.org: no such host  app=tarmak
DEBU[0000] trying to connect to https://api.workshop-cluster.tarmak.org  app=tarmak
WARN[0000] error connecting to cluster: Get https://api.workshop-cluster.tarmak.org/version: dial tcp: lookup api.workshop-cluster.tarmak.org: no such host  app=tarmak
DEBU[0000] trying to connect to https://api.workshop-cluster.tarmak.org  app=tarmak
WARN[0000] error connecting to cluster: Get https://api.workshop-cluster.tarmak.org/version: dial tcp: lookup api.workshop-cluster.tarmak.org: no such host  app=tarmak
DEBU[0000] trying to connect to https://api.workshop-cluster.tarmak.org  app=tarmak
WARN[0000] error connecting to cluster: Get https://api.workshop-cluster.tarmak.org/version: dial tcp: lookup api.workshop-cluster.tarmak.org: no such host  app=tarmak
KUBECONFIG=/Users/luke/go/src/github.com/jetstack/advanced-workshop/tarmak/test-luke/workshop-cluster/kubeconfig
```
Even though it errors it still echos `KUBECONFIG`. 

I think it'd also be useful to be able to specify the destination file rather than the default that could be overwritten with `tarmak clusters apply`

MattiasGees:
@dippynark Indeed about the custom path, that is a good idea and I will add it.

I think for the validity of the SSL cert, we need to use the domain name we now use. As far as I know, we can't use the IP address or ELB cname as those are not valid in the SSL cert. Or am I missing something?

dippynark:
@MattiasGees I don't think we need to validate the SSL cert as part of the kubeconfig command. All we need to return from the kubeconfig command is the kubeconfig we return already but with the `server:` part replaced by `api.workshop-cluster.tarmak.org` (when using the public endpoint) which can be found by querying AWS (in Route53) or using terraform state, whichever is more appropriate.

If everything is correct, the SSL cert should be valid.

MattiasGees:
@dippynark I think it is not bad to check if our kubeconfig works before we actually give the user the kubeconfig. That way we give a valid error to the user if the connection can't be me made. By keeping this, we are also inline with the `kubectl` command which has the same way of working of first testing the connection before actually executing the `kubectl` command.

What I find weird, is that you didn't get an error but got the kubeconfig as output. I did the same test with a wrong url and I got an full error.

```
DEBU[0001] trying to connect to https://api2.singlecluster-cluster.develop.tarmak.org  app=tarmak
WARN[0001] error connecting to cluster: Get https://api2.singlecluster-cluster.develop.tarmak.org/version: dial tcp: lookup api2.singlecluster-cluster.develop.tarmak.org: no such host  app=tarmak
DEBU[0001] trying to connect to https://api2.singlecluster-cluster.develop.tarmak.org  app=tarmak
WARN[0001] error connecting to cluster: Get https://api2.singlecluster-cluster.develop.tarmak.org/version: dial tcp: lookup api2.singlecluster-cluster.develop.tarmak.org: no such host  app=tarmak
DEBU[0001] trying to connect to https://api2.singlecluster-cluster.develop.tarmak.org  app=tarmak
WARN[0001] error connecting to cluster: Get https://api2.singlecluster-cluster.develop.tarmak.org/version: dial tcp: lookup api2.singlecluster-cluster.develop.tarmak.org: no such host  app=tarmak
DEBU[0001] trying to connect to https://api2.singlecluster-cluster.develop.tarmak.org  app=tarmak
WARN[0001] error connecting to cluster: Get https://api2.singlecluster-cluster.develop.tarmak.org/version: dial tcp: lookup api2.singlecluster-cluster.develop.tarmak.org: no such host  app=tarmak
DEBU[0001] trying to connect to https://api2.singlecluster-cluster.develop.tarmak.org  app=tarmak
WARN[0001] error connecting to cluster: Get https://api2.singlecluster-cluster.develop.tarmak.org/version: dial tcp: lookup api2.singlecluster-cluster.develop.tarmak.org: no such host  app=tarmak
FATA[0001] unable to connect to kubernetes after 5 tries
```

simonswine:
This needs some refactoring/rebasing and the dco stuff

/unassign
/assign @MattiasGees

MattiasGees:
All fixed now.
/unassign
/assign @simonswine

simonswine:
This autodetect in my comment earlier (https://github.com/jetstack/tarmak/pull/328#issuecomment-399094135). Is still not working. I am on a single cluster without public API

```
tarmak cluster kubeconfig
FATA[0001] APIServer is not public and we are not able to connect to the APIserver without SSH tunnel, please use tarmak cluster kubectl
```

/assign @MattiasGees 
/unassgin

MattiasGees:
Fixed now and found another small bug that is fixed now.
/unassign
/assign @simonswine

JoshVanL:
What is the status of this? Can I help move it along for 0.5 ? @simonswine @MattiasGees

simonswine:
Closing this in favour of #632

PR #331, opened 2018-06-21
Adds validation to enforce single non-worker instance pools

### LINKED ISSUE #320: Enfore single instance pool for all types except worker

/kind bug

**What happened**: I can create multiple master/etcd/basition/vaul/jenkins instance pools in a single cluster. This doesn't make sense

**What you expected to happen**: Tarmak should error early (not just due to terraform templating) when creating multiple of the above instance pools

**How to reproduce it (as minimally and precisely as possible)**: create multiple of any of the above instance pools

JoshVanL:
Currently tarmak does not error early when creating multiple non-worker instance pools. This adds a validation check to top this.

fixes #320 

**Release note**:

```release-note
NONE
```

/cc @dippynark

### DIFF
--- a/pkg/tarmak/cluster/cluster.go
+++ b/pkg/tarmak/cluster/cluster.go
@@ -218,12 +218,12 @@ func (c *Cluster) validateInstancePoolMultiple() error {
 
 	list := make(map[string][]*clusterv1alpha1.InstancePool)
 	for _, i := range c.Config().InstancePools {
-		list[i.Name] = append(list[i.Name], &i)
+		list[i.Type] = append(list[i.Type], &i)
 	}
 
-	for name, v := range list {
-		if name != "worker" && len(v) > 1 {
-			err := fmt.Errorf("'%s' only supports a single instance pool", name)
+	for instanceType, v := range list {
+		if instanceType != clusterv1alpha1.InstancePoolTypeWorker && len(v) > 1 {
+			err := fmt.Errorf("instance pool type '%s' only supports a single instance pool", instanceType)
 			result = multierror.Append(result, err)
 		}
 	}
--- a/pkg/tarmak/cluster/cluster.go
+++ b/pkg/tarmak/cluster/cluster.go
@@ -178,7 +178,9 @@ func (c *Cluster) VerifyInstancePools() (result error) {
 	return nil
 }
 
-func (c *Cluster) Validate() (result error) {
+func (c *Cluster) Validate() error {
+	var result *multierror.Error
+
 	// validate instance pools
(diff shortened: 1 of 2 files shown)


pkg/tarmak/cluster/cluster.go
```diff
@@ -203,7 +205,30 @@ func (c *Cluster) Validate() (result error) {
...
+
+	return result.ErrorOrNil()
+}
+
+func (c *Cluster) validateInstancePoolMultiple() error {
+	var result *multierror.Error
+
+	list := make(map[string][]*clusterv1alpha1.InstancePool)
+	for _, i := range c.Config().InstancePools {
+		list[i.Name] = append(list[i.Name], &i)
```
dippynark: should use Type here instance of Name - otherwise having two instance pools of type jenkins called jenkins1 and jenkins2 still works

pkg/tarmak/cluster/cluster.go
```diff
@@ -203,7 +205,30 @@ func (c *Cluster) Validate() (result error) {
...
+	}
+
+	for name, v := range list {
```
dippynark: name -> type

pkg/tarmak/cluster/cluster.go
```diff
@@ -203,7 +205,30 @@ func (c *Cluster) Validate() (result error) {
...
+		if name != "worker" && len(v) > 1 {
```
dippynark: "worker" -> clusterv1alpha1.InstancePoolTypeWorker

pkg/tarmak/cluster/cluster.go
```diff
@@ -203,7 +205,30 @@ func (c *Cluster) Validate() (result error) {
...
+			err := fmt.Errorf("'%s' only supports a single instance pool", name)
```
dippynark: Maybe change message to "Instance pool type '%s' only supports a single instance pool"

pkg/tarmak/cluster/cluster.go
JoshVanL: done

pkg/tarmak/cluster/cluster.go
JoshVanL: done

pkg/tarmak/cluster/cluster.go
JoshVanL: done

pkg/tarmak/cluster/cluster.go
JoshVanL: done

JoshVanL:
Consolidating into https://github.com/jetstack/tarmak/pull/347

PR #335, opened 2018-06-21
Multi-cluster environment  instance pool validation check

### LINKED ISSUE #333: Fail when wrong instance types specified in multi-cluster environment

/kind bug

**What happened**: In a multi-cluster environment, hub related instance pools can be put in the kubernetes cluster and vice versa. Tarmak won't create the instance pool but it is still wrong configuration.

**What you expected to happen**: Tarmak should error if there is a bastion instance pool in the kubernetes cluster and a master instance pool in the hub (for example)

**How to reproduce it (as minimally and precisely as possible)**: Place a bastion instance pool in a kubernetes clusters in a multi-cluster environment

JoshVanL:
This adds a validation check to ensure instance pool types in multi-cluster environments are correct according to their cluster.

**Which issue this PR fixes**
fixes #333 

**Release note**:

```release-note
NONE
```

/cc @dippynark

### DIFF
--- a/pkg/tarmak/cluster/cluster.go
+++ b/pkg/tarmak/cluster/cluster.go
@@ -203,7 +203,54 @@ func (c *Cluster) Validate() (result error) {
 		}
 	}
 
+	// validate instance instance pools according to hub and kubernetes multi cluster
+	if err := c.validateMultiClusterInstancePoolTypes(); err != nil {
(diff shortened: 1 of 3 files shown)


JoshVanL:
Consolidating into https://github.com/jetstack/tarmak/pull/347

PR #345, opened 2018-06-23
Upgrade Tiller puppet module and docs

dippynark:
**What this PR does / why we need it**: Upgrade the Tiller puppet module and docs

```release-note
Upgrade Tiller manifest versions and documentation
```

### DIFF
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -365,27 +365,34 @@ certificate is valid for ``jenkins.<environment>.<zone>``.
 Tiller
 ~~~~~~
 
-Another configuration option allows to deploy Tiller the server-side of `Helm
-<https://github.com/kubernetes/helm>`_. Tiller is listening for request on the
-loopback device only. This makes sure that no other Pod in the cluster can
-speak to it, while Helm clients are still able to access it using a port
-forwarding through the API server.
-
-As Helm and Tiller minor version need to match, the tarmak configuration also
-allows to override the deployed version:
+Tarmak supports deploying Tiller, the server-side component of `Helm
+<https://github.com/kubernetes/helm>`_, when spinning up a Kubernetes cluster.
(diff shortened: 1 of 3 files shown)


puppet/modules/kubernetes_addons/manifests/tiller.pp
```diff
@@ -1,10 +1,40 @@
...
+    $_image = 'gcr.io/kubernetes-helm/tiller'
+  } else {
+    $_image = $image
+  }
+
+  if $version == undef {
+    $_version = '2.9.1'
+  } else {
+    $_version = $version
+  }
```
simonswine: Why do you not keep it as a parameter default?

puppet/modules/kubernetes_addons/manifests/tiller.pp
```diff
@@ -1,10 +1,40 @@
...
+  Optional[String] $version=undef,
   String $namespace='kube-system',
 ) inherits ::kubernetes_addons::params {
   require ::kubernetes
 
+  if $image == undef {
+    $_image = 'gcr.io/kubernetes-helm/tiller'
+  } else {
+    $_image = $image
+  }
```
simonswine: Why do you not keep it as a parameter default

docs/user-guide.rst
```diff
@@ -365,27 +365,34 @@ certificate is valid for ``jenkins.<environment>.<zone>``.
...
+in the cluster connecting to its unauthenticated endpoint. Helm clients can
+still talk to Tiller by port forwarding through the Kubernetes API Server. The
+following `tarmak.yaml` snippet shows how you would enable Tiller.
 
 .. code-block:: yaml
 
-  kubernetes:
-    tiller:
-      enabled: true
-      version: 2.9.1
```
simonswine: Can we please keep that example how to change the version/tag of tiller

simonswine:
And also some better release-notes would have been great

PR #356, opened 2018-06-29
Forwarding Interrupt to processes

### LINKED ISSUE #169: Ensure CTRL-C is forwarded correctly to packer/terraform

/kind bug

**What happened**:

We don't really handle CTRL-C events probably. 

What we should do: Ensure the signal gets forwarded to the subprocess and has a chance to exit cleanly

`tarmak cluster plan` and press CTRL-C

JoshVanL:
Currently we do not forward signal interrupts very well to sub processes and running tasks such at terraform and packer. This PR properly handles singles and forwards them to each running task, gracefully exiting.

 fixes #169  

```release-note
Propagate interrupt signals to sub-processes and tasks
```


JoshVanL:
@simonswine since the removing of the remote state lock is quite dependant on the context being propagated better, I'm going to have that 'rebased' on top of the kill functionality if that is OK.

simonswine:
I think this needs more work. We should directly forward signals and wait for subprocess to exit and then exit tarmak.

Currently we still have the workaround that asks to send another signal before doing anything.

I would say that we should not worry about the containerised packer case for now. We should invest the time into that ticket instead #362. And integrate packer into tarmak directly as we did that with terraform.

Task as I see them:

- [ ] Use a signal channel to forward the exact same signal received. (We still should have a white list for signals to forward, should be SIGINT and SIGTERM for now)
- [ ] Act on the first signal, wait for subprocess to finish. (waitGroup)

/unassign
/assign @JoshVanL

simonswine:
Will wait for #372 to be merged first 

/hold

pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -63,6 +63,7 @@ type EC2 interface {
 type DynamoDB interface {
 	DescribeTable(input *dynamodb.DescribeTableInput) (*dynamodb.DescribeTableOutput, error)
 	CreateTable(input *dynamodb.CreateTableInput) (*dynamodb.CreateTableOutput, error)
+	GetItem(input *dynamodb.GetItemInput) (*dynamodb.GetItemOutput, error)
```
simonswine: I think we are not using that or am I missing something

pkg/tarmak/interfaces/interfaces.go
```diff
@@ -271,3 +273,10 @@ type Initialize interface {
 	CurrentProvider() Provider
 	CurrentEnvironment() Environment
 }
+
+type Context interface {
```
simonswine: Not too happy with calling that Context. This name is quite overloaded (Context from Kubeconfig file, golang's `import "context"`, ...

Maybe we can call it `CancelationContext`?

pkg/tarmak/provider/amazon/amazon.go
JoshVanL: done

pkg/tarmak/interfaces/interfaces.go
JoshVanL: done

simonswine:
Thanks for implementing that @JoshVanL work well. Sorry about the confusion packer sometimes cleans up really quickly sometimes it takes quite a while

/approve
/lgtm

PR #360, opened 2018-07-09
Remove terraform remote state lock

JoshVanL:
Terraform does not remove it's own state lock on interrupt. This PR attempts to remove this remote state lock during an interrupt of a terraform oporation.

Rebased on top of https://github.com/jetstack/tarmak/pull/356

**Release note**:

```release-note
Remove terraform remote state lock during an operation interrupt
```


simonswine:
The CLI terraform build removes locks after it receives SIGINT/SIGTERM we should investigate why the tarmak builtin terraform is not behaving like that and if we correctly forward the the signal to the process. I think we should not implement specific terraform controlled operations. (Like writing to the dynamo lock db of terraform)

Can you please investigate more why terraform is misbehaving @JoshVanL 

/unassign 
/assign @JoshVanL

JoshVanL:
Achieved via #356 
/close

PR #367, opened 2018-07-13
Upgrade kubernetes versions for testing

simonswine:
```release-note
NONE
```


puppet/modules/tarmak/Makefile
```diff
@@ -1,46 +1,66 @@
 BUNDLE_DIR ?= .bundle
 
+VERSION_1_10 := 1.11.0
```
MattiasGees: Shouldn't this be 1_11?

simonswine: Good point!

Issue #368, opened 2018-07-16
Add labels & taints option to instance pool config


charlieegan3:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
It is not possible to add Taints or Labels to nodes in Tarmak instance pools from tarmak.yaml.

**What you expected to happen**:
I'd expect to be able to configure it something like this:

```
  instancePools:
  - image: centos-puppet-agent
    maxCount: 1
    metadata:
      creationTimestamp: "2018-05-09T11:27:35Z"
      name: worker
    minCount: 3
    ***
    labels:
    - key: label_name
      value: label_value
    taints:
    - key: taint_name
      value: taint_value
      effect: taint_effect
    ***
    size: large
    type: worker
    ...
```

**Anything else we need to know?**:

We already have this in the puppet config here: https://github.com/jetstack/tarmak/tree/master/puppet/modules/kubernetes#node_labels

We also appear to have taints down as a kubelet param here: https://github.com/jetstack/tarmak/blob/master/puppet/modules/kubernetes/templates/kubelet.service.erb#L17

simonswine:
Tanks this sounds good. Only thing I found is `taints` needs a `:`

charlieegan3:
/close

Completed in #402

PR #369, opened 2018-07-17
 Add labels & taints option to instance pool config

charlieegan3:
Enables the setting of labels and taints on nodes via kubelet flags in the tarmal.yaml

**Which issue this PR fixes**

https://github.com/jetstack/tarmak/issues/368

**Release note**:
```release-note
Enable the setting of node labels and taints via kubelet flags
```

### DIFF
--- a/pkg/apis/cluster/v1alpha1/label.go
+++ b/pkg/apis/cluster/v1alpha1/label.go
@@ -14,15 +14,8 @@
 
 package v1alpha1
 
-import (
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-)
-
 // Label structure for instancepool node labels
 type Label struct {
-	metav1.TypeMeta   `json:",inline"`
-	metav1.ObjectMeta `json:"metadata,omitempty"`
-
 	Key   string `json:"key,omitempty"`
 	Value string `json:"value,omitempty"`
 }
--- a/pkg/apis/cluster/v1alpha1/label.go
+++ b/pkg/apis/cluster/v1alpha1/label.go
@@ -1,5 +1,4 @@
 // Copyright Jetstack Ltd. See LICENSE for details.
-// Copyright © 2017 The Kubicorn Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/apis/cluster/v1alpha1/label.go
+++ b/pkg/apis/cluster/v1alpha1/label.go
@@ -1,4 +1,5 @@
 // Copyright Jetstack Ltd. See LICENSE for details.
+// Copyright © 2017 The Kubicorn Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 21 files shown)


pkg/puppet/puppet.go
```diff
@@ -354,6 +354,13 @@ func (p *Puppet) writeHieraData(puppetPath string, cluster interfaces.Cluster) e
 			variables = append(variables, fmt.Sprintf(`kubernetes_addons::cluster_autoscaler::instance_pool_name: "%s"`, workerInstancePoolName))
 		}
 
+		if len(instancePool.Config().Labels) > 0 {
+			variables = append(variables, fmt.Sprintf("kubernetes::kubelet::node_labels:\n%s", instancePool.Labels()))
+		}
+		if len(instancePool.Config().Taints) > 0 {
+			variables = append(variables, fmt.Sprintf("kubernetes::kubelet::node_taints:\n%s", instancePool.Taints()))
+		}
```
simonswine: I think this will override node-taints and labels we currently set here:

https://github.com/jetstack/tarmak/blob/master/puppet/modules/kubernetes/manifests/kubelet.pp#L112

We need to make sure that master nodes are still label and taint like that and that worker nodes are actually labeld with role="worker".

We should only be able to override that by using the same key in the tarmak.yaml

This seems to be rather complex so ideally I want to test those cases.

pkg/apis/cluster/v1alpha1/label.go
```diff
@@ -0,0 +1,29 @@
+// Copyright Jetstack Ltd. See LICENSE for details.
+// Copyright © 2017 The Kubicorn Authors
```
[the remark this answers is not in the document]
charlieegan3: copy paste job, as usual :wink: 

Removed from all the other files too in 9d0d908c

pkg/puppet/puppet.go
charlieegan3: How about something like this? https://github.com/jetstack/tarmak/pull/369/commits/60511e45148c0f5f8320de30b867856e993c3933

munnerz: (drive-by review): how is variable/string escaping handled? What if I specify a taint with `:` in it or something?

pkg/apis/cluster/v1alpha1/label.go
```diff
@@ -0,0 +1,28 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Label structure for instancepool node labels
+type Label struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
munnerz: How come TypeMeta and ObjectMeta? Do labels get named/labels/annotations/kinds etc? (same applies for Taints)

charlieegan3: I'll confess that this is just coming from [Volume](https://github.com/jetstack/tarmak/blob/master/pkg/apis/cluster/v1alpha1/volume.go#L30) and probably doesn't need to be here :flushed:

charlieegan3: Fixed in 1fab9f97

pkg/puppet/puppet.go
charlieegan3: I've had a go at making this a little more robust & added some better error reporting in passing too. https://github.com/jetstack/tarmak/pull/369/commits/d150d9ed92aaf03f99803d08a8ab6c598bd83bce

simonswine:
/assign @charlieegan3 
/unassign

Some more docs would be great :slightly_smiling_face:

PR #371, opened 2018-07-17
Revert "vault-helper dry run and delete"

simonswine:
Tempted to roll that back for now

:grimacing: 

Any risks with that @JoshVanL?


JoshVanL:
@simonswine 

I agree, should be okay to do that

PR #372, opened 2018-07-17
Integrate packer into tarmak binary

### LINKED ISSUE #362: Integrate packer into tarmak
/kind feature

**What happened**:

Traditionally we integrated terraform and packer using docker. For terraform we moved to the model of integrating the go code directly into tarmak. The only thing that is run like that currently is Packer.

I would consider that as technical debt and would like to get rid of the requirement of a local installed docker.

Packer is built into the tarmak binary

JoshVanL:
Integrate packer into the Tarmak binary instead of using docker to apply image build.

fixes #362 

**Special notes**:
Uses a modified fork of the packer package.

```release-note
Removes Docker dependency
```


JoshVanL:
/assign @simonswine 
@simonswine 

In order to resolve the dependencies I have created a fork of packer and removed the troublesome packages - some builders which aren't necessary. Due to the way that packer has been written it is forcing a lot of extra dependencies to be imported which are not needed - the PR is adding 6x the lines of tarmak currently. I think either I should prune more builders which aren't required from the fork or we should think about using the packer binary instead. We could have an 'install' of tarmak where /opt/tamark keeps a packer binary and maybe terraform too. The first run of the tarmak would pull the dependant binaries and set up /opt/tarmak.
What do you think?

simonswine:
/assign @JoshVanL 
/unassign

Need rebase I guess

simonswine:
/assign @JoshVanL
/unassign

Can you look into that failing generations?
```
make clean verify
```
Reproduces it for me

JoshVanL:
@simonswine 
There were two trouble packages - mock and gengo.
I have pinned mock to 1.0.1 which is what we were using before instead of the latest 1.1.1
This is was causing problems where generated mock files were now importing packages with the vendor directory, seems to be a bug in the latest releases.
Gengo has no releases but the master branch seems so have some incompatibility. It's now pinned to the same commit as Tarmak's master.

These failing tests I'm not able to reproduce yet.

pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -603,3 +603,22 @@ func (a *Amazon) VolumeType(typeIn string) (typeOut string, err error) {
...
+		return nil, fmt.Errorf("failed to get amazon credentials: %v", err)
+	}
+
+	return map[string]string{
+		"AWS_ACCESS_KEY_ID":     creds.AccessKeyID,
+		"AWS_SECRET_ACCESS_KEY": creds.SecretAccessKey,
+		"AWS_SESSION_TOKEN":     creds.SessionToken,
+		"AWS_DEFAULT_REGION":    a.Region(),
+	}, nil
+}
```
simonswine: Can we use https://github.com/JoshVanL/tarmak/blob/389e38813a090c1659b86ec582f228d1de5df587/pkg/tarmak/provider/amazon/amazon.go#L286 instead?

JoshVanL: Yes sure

PR #382, opened 2018-07-25
Move consul and puppet cloud-init into puppet

### LINKED ISSUE #93: Use puppet for the Consul/Vault instances
Currently the vault/consul instances are spun up using static cloud-init. Tarmak should use puppet modules to setup consul/vault clusters

JoshVanL:
Moves consul and vault install into puppet modules.

fixes #93

**Release note**:

```release-note
NONE
```


simonswine:
I think I got pretty for with #98, hope you can make use of that.

The vault module should be IMHO completely separate from the existing vault_client one

puppet/modules/consul/manifests/config.pp
```diff
@@ -0,0 +1,110 @@
...
+    }
+    -> file { $consul::config_path:
+        ensure  => file,
+        owner   => $consul::user,
+        group   => $consul::group,
+        mode    => '0600',
+        content => template('consul/consul.json.erb'),
+    }
+
+    file { "${consul::vault_config_dir}/vault.hcl":
```
simonswine: Why are we writing vault config from consul module?

puppet/modules/airworthy/manifests/install.pp
```diff
@@ -35,4 +35,9 @@
     require => File[$::airworthy::bin_dir],
   }
 
+  file { '/bin/airworthy':
```
simonswine: Why is this necessay? We should not modify /usr/** /bin or any of those package managed directories

puppet/modules/consul/manifests/config.pp
```diff
@@ -0,0 +1,110 @@
...
+            default => $consul::retry_join,
+        },
+        ca_file             => $consul::ca_file,
+        cert_file           => $consul::cert_file,
+        key_file            => $consul::key_file,
+        verify_outgoing     => $enable_tls,
+        verify_incoming     => $enable_tls,
+    }
+
+    ## TODO: Add AWS stuff for rejoining
```
simonswine: We really need that for the first iteration, or we have a way utilising a DNS record as well

puppet/modules/consul/manifests/init.pp
```diff
@@ -0,0 +1,108 @@
...
+    $environment,
+    $backup_bucket_prefix,
+    $backup_schedule,
+    $volume_id,
+
+    String $app_name = $consul::params::app_name,
+    String $version = $consul::params::version,
+    String $bin_dir = $consul::params::bin_dir,
+    String $config_dir = $consul::params::config_dir,
+    String $vault_config_dir = $consul::params::vault_config_dir,
```
simonswine: This is should not be here

puppet/modules/consul/manifests/params.pp
```diff
@@ -0,0 +1,23 @@
+# == Class consul::params
+#
+# This class is meant to be called from consul.
+# It sets variables according to platform.
+#
+class consul::params {
+    $app_name = 'consul'
+    $version = '1.2.1'
+    $bin_dir = '/usr/local/bin'
+    $vault_config_dir = '/etc/vault'
```
simonswine: This should not be here

simonswine:
/assign @JoshVanL 
/unassign

I had a look at the consul implementation. See my comments. We should not use /usr/local for installing anything in tarmak. It is ok to install things for testing in there (the vault server for acceptance tests for example)

Once consul looks great I would like to move on to the vault module

PR #391, opened 2018-07-31
Ensures results are of type *multierror.Error and call ErrorOrNil at return

JoshVanL:
Due to a result being miss returned a bug was created. This PR ensures all append results are of *multierror.Error and return ErrorOrNil()

```release-note
NONE
```


simonswine:
Not too sure why we need that?! This is exactly as usage suggest https://github.com/hashicorp/go-multierror

PR #395, opened 2018-08-01
Make vault-helper download URL customizable

### LINKED ISSUE #393: Customize vault-helper download URL

/kind feature

It would be useful to allow the override of the `vault-helper` download url in vault_client module so it can be downloaded from local mirrors instead of github.com

Wiston999:
**What this PR does / why we need it**: This PR lets vault-helper download URL to be modifiable to point to internal mirrors.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #393 

**Release note**:

```release-note
```

### DIFF
--- a/puppet/modules/vault_client/manifests/init.pp
+++ b/puppet/modules/vault_client/manifests/init.pp
@@ -25,6 +25,7 @@ class vault_client (
   $version = $::vault_client::params::version,
   $bin_dir = $::vault_client::params::bin_dir,
   $download_dir = $::vault_client::params::download_dir,
+  $download_url = $::vault_client::params::download_url,
   $dest_dir = $::vault_client::params::dest_dir,
   $server_url = $::vault_client::params::server_url,
   $systemd_dir = $::vault_client::params::systemd_dir,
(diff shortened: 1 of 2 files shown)


simonswine:
I think we need documetation for that and also provide similar things for other downloads we do:
- hyperkube
- vault-helper
- wing

Wiston999:
Hi sorry for the delay on the update.
I've added the same capability for hyperkube and also reflected those parameters on README files for both modules.

I couldn't find anything related to wing in the puppet/ subfolder so I haven't make any change related to it.

JoshVanL:
Hi @Wiston999, thanks for the PR. Although I have also created this PR https://github.com/jetstack/tarmak/pull/420 that documents how to customize various URLs in puppet and terraform, I think that it might make more sense to have this as configuration options in the tarmak.yaml . This would prevent having to recompile and provide a much better UX IMO. This can be implemented much like this PR https://github.com/jetstack/tarmak/pull/439 but can also be propagated to terraform to set the wing URL too. Thoughts @simonswine ?

PR #396, opened 2018-08-01
Recover properly when bastion goes down

### LINKED ISSUE #187: Cannot recover from failed bastion instance
/kind bug

**What happened**: The bastion instance went down (accidentally deleted the public route) and running `tarmak clusters apply` hangs checking whether the bastion is up.

**What you expected to happen**: For `tarmak clusters apply` to bring the bastion back up

**How to reproduce it (as minimally and precisely as possible)**: Create a hub and delete the bastion

dippynark:
**What this PR does / why we need it**: Allows the bastion to be created by reapplying when there it goes down

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #187

```release-note
Allow bastion to be recreated by reapplying
```

### DIFF
--- a/pkg/terraform/providers/tarmak/data_source_bastion_instance.go
+++ b/pkg/terraform/providers/tarmak/data_source_bastion_instance.go
@@ -19,10 +19,6 @@ func dataSourceBastionInstance() *schema.Resource {
 				Type:     schema.TypeString,
 				Required: true,
 			},
-			"instance_id": {
-				Type:     schema.TypeString,
-				Optional: true,
-			},
 			"username": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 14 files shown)


simonswine:
I need a bit more context what this is trying to achieve

dippynark:
@simonswine at the time this was written, the bastion would not be recreated from a reapply if it was deleted or was down for whatever reason - it also tries to speed up plan time when the bastion is down (so that the tarmak vault cluster resource and vault token resources don't hang for ages) but due to issues with terraform not rereading data resources at refresh time (https://github.com/hashicorp/terraform/issues/17034) this was not perfect

Issue #400, opened 2018-08-01
Relax label and taints validations


simonswine:
As part of #369 we merged a rather strict validation of taints and labels.

I suggest we should make sure we follow the upstream validations for those objects (e.g. labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

I think the the upstream code for taints is not really re-usable, but we should just allow empty values and a similar character set:

- Tolerations: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/validation/validation.go#L2826

Labels should be quite straightforward: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/validation/validation.go#L66

/assign @charlieegan3 
/kind bug

PR #402, opened 2018-08-02
Relax Label and Taint validations

charlieegan3:
- Taints and labels can have blank values
- Keys can have slashes and other characters as permitted here:
  https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

Fixes https://github.com/jetstack/tarmak/issues/400

I haven't checked that this is templated correctly by the puppet code yet

**Release note**:
```release-note
Allow blank values and a wider range of keys for instance pool labels & taints
```

### DIFF
--- a/pkg/tarmak/instance_pool/instance_pool.go
+++ b/pkg/tarmak/instance_pool/instance_pool.go
@@ -206,45 +206,71 @@ func (n *InstancePool) ValidateAllowCIDRs() (result error) {
 
 func (n *InstancePool) Labels() (string, error) {
 	var labels []string
+
+	err := n.validLabels()
+	if err != nil {
+		return "", err
+	}
+
+	for _, label := range n.conf.Labels {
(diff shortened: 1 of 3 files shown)


pkg/tarmak/instance_pool/instance_pool.go
```diff
@@ -206,45 +206,71 @@ func (n *InstancePool) ValidateAllowCIDRs() (result error) {
...
+	}
+
+	for _, label := range n.conf.Labels {
+		labels = append(labels, fmt.Sprintf("  %s: \"%s\"", label.Key, label.Value))
+	}
+
+	return strings.Join(labels, "\n"), nil
+}
+
+func (n *InstancePool) validLabels() error {
```
simonswine: Can we use that instead: https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/validation/validation.go#L66

PR #439, opened 2018-08-09
Expose feature flags

### LINKED ISSUE #431: Expose feature gates in tarmak.yaml
We should expose feature gates for all components involved

/kind feature

JoshVanL:
Adds configuration options to enable/disable feature gates on Kubernetes components

fixes #431 

```release-note
Adds feature gates options to Tarmak configuration
```

/assign @simonswine


docs/user-guide.rst
```diff
@@ -757,6 +757,30 @@ make sure you don't define any volumes with the name ``docker``.
...
+Tarmak configuration file. The feature gates can be set on the API server,
+Kubelet, Scheduler, Controller Manager and Kube-Proxy that will take effect cluster
+wide. To change feature gates, each component takes a list of stings that
+will be applied to their corresponding command line flags or configuration
+file under the Kubernetes code block like following:
+
+.. code-block:: yaml
+
+    kubernetes:
+      featureGates:
```
MattiasGees: @JoshVanL I would like to see the same config style as we already have for admission controller, public elb. 
This would mean we do `kubernetes -> apiServer -> featureGates` instead of `kubernetes -> featureGates -> apiServer`.

A full example:

```
  kubernetes:
    apiServer:
      featureGates:
        - "AllAlpha=true"
        - "APIResponseCompression=false"
    kubelet:
      featureGates:
        - "CustomPodDNS=true"
    kubeProxy:
      featureGates:
        - "CSIPersistentVolume=true"
    controllerManager:
    scheduler:
```


In a full example this would become:

```
  kubernetes:
    apiServer:
      featureGates:
        - "AllAlpha=true"
        - "APIResponseCompression=false"
      enableAdmissionControllers:
        - "DefaultStorageClass"
      disableAdmissionControllers:
        - "DefaultTolerationSeconds"
      public: true
```

This will remove a lot of the confusion.

JoshVanL: :+1:

pkg/puppet/puppet.go
```diff
@@ -154,6 +154,22 @@ func kubernetesClusterConfig(conf *clusterv1alpha1.ClusterKubernetes, hieraData
 		hieraData.classes = append(hieraData.classes, `prometheus`)
 	}
 
+	if a := conf.APIServer; a != nil && len(a.FeatureGates) > 0 {
+		hieraData.variables = append(hieraData.variables, fmt.Sprintf(`kubernetes::api_feature_gates: ["%s"]`, strings.Join(a.FeatureGates, `", "`)))
```
MattiasGees: why not use `kubernetes::apiserver::feature_gates` as var and only define the var in the correct class? We don't need that variable in the `apiserver` class or any other subclass of it.

I have the same remark for the other components.

puppet/modules/calico/manifests/init.pp
```diff
@@ -10,6 +10,7 @@
   String $namespace = 'kube-system',
   Optional[String] $pod_network = undef,
   Integer[1000,65535] $mtu = 1480,
+  Array[String] $feature_flags = [],
```
MattiasGees: Is this used?

puppet/modules/kubernetes/manifests/controller_manager.pp
```diff
@@ -28,6 +29,14 @@
     $version_before_1_6 = true
   }
 
+  if $::kubernetes::controller_feature_gates == [] {
+    $_feature_gates = delete_undef_values([
+      $::kubernetes::_enable_pod_priority ? { true => 'PodPriority=true', default => undef },
```
MattiasGees: I don't think this is needed. Only needed for `API server, scheduler and kubelet` https://v1-10.docs.kubernetes.io/docs/concepts/configuration/pod-priority-preemption/

puppet/modules/kubernetes/manifests/proxy.pp
```diff
@@ -23,6 +24,14 @@
     ensure => 'present',
   })
 
+  if $::kubernetes::proxy_feature_gates == [] {
+    $_feature_gates = delete_undef_values([
+      $::kubernetes::_enable_pod_priority ? { true => 'PodPriority=true', default => undef },
```
MattiasGees: I don't think this is needed. Only needed for API server, scheduler and kubelet https://v1-10.docs.kubernetes.io/docs/concepts/configuration/pod-priority-preemption/

MattiasGees:
Thanks, tested and verified. Everything works as expected.

/lgtm
/approve
/unassign
/assign @simonswine

simonswine:
This needs some proper refactoring after #442 merged, moving it out of milestone

/assign @JoshVanL 
/unassign @MattiasGees 
/unassign @simonswine

pkg/puppet/puppet.go
```diff
@@ -165,9 +166,72 @@ func kubernetesClusterConfig(conf *clusterv1alpha1.ClusterKubernetes, hieraData
 		hieraData.classes = append(hieraData.classes, `prometheus`)
 	}
 
+	globalGates := make(map[string]bool)
+	if conf.GlobalFeatureGates != nil {
+		globalGates = conf.GlobalFeatureGates
+	}
+
+	if a := conf.APIServer; a != nil {
```
simonswine: This won't set global feature gate, if no APIserver specific gates are set

pkg/puppet/puppet.go
```diff
@@ -165,9 +166,72 @@ func kubernetesClusterConfig(conf *clusterv1alpha1.ClusterKubernetes, hieraData
...
+		// pod prio
```
simonswine: this comment is making too much sense

pkg/puppet/puppet.go
simonswine: Maybe check the length of the featureGatesString return

simonswine:
/approve 
/lgtm

Thanks you @JoshVanL this works for me

PR #443, opened 2018-08-13
Copy docker files even when start fails

JoshVanL:
This ensures that even if `docker start` fails when trying to build the docs, the output build files are copied out. This helps to get build error messages.

```release-note
NONE
```

/assign @simonswine


docs/Makefile
```diff
@@ -32,7 +32,7 @@ docker_html:
 	# start documentaion build
 	$(eval CONTAINER_ID := $(shell docker create $(IMAGE_ID)))
 	(git ls-files && git ls-files --others --exclude-standard) | tar cf -  -T - | docker cp - $(CONTAINER_ID):/site
-	docker start -a $(CONTAINER_ID)
+	- docker start -a $(CONTAINER_ID)
```
simonswine: This will ignore failing doc builds, you need to store the exit code and return it after the docker cp

PR #446, opened 2018-08-16
WIP: Default to public pre-built AMI images if none exist

### LINKED ISSUE #415: Provide prebuilt AMI image per region
**What happened**:

We currently rely on the user to build their packer images. This takes a while and will potentially a bad first impression with new users

I would like to have some AMI published with every release that are used by default, if no local AMI is found.

JoshVanL:
Default to pre-built AMI images if the user has not build them yet.

fixes #415 

```release-note
Provide pre-built AMI images to fallback to 
```

### DIFF
--- a/pkg/packer/image.go
+++ b/pkg/packer/image.go
@@ -92,7 +92,7 @@ func (i *image) Build() (amiID string, err error) {
 
 	envVars, err := i.tarmak.Provider().Environment()
 	if err != nil {
-		return "", fmt.Errorf("faild to get provider credentials: %v", err)
+		return "", fmt.Errorf("failed to get provider credentials: %v", err)
 	}
 
 	var result *multierror.Error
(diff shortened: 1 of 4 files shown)


JoshVanL:
@simonswine The current default AMI for Tarmak uses a non-free image so the resulting AMI contains a product code. This prevent's these AMIs being made public. What do you think should be used instead?

simonswine:
Instead of relying of built-in ec2 AMI IDs, we should get the latest image using the API

```
aws ec2 describe-images --region eu-central-1 --owners 344758251446 --filters Name=name,Values=Tarmak\ 0\.6*
```

The values should come from Major and Minor (Minor needs to replace the + postfix) in `pkg/version`

Then we should look into the tags `tarmak_version` and  `tarmak_base_image_name` of the images and compare with the required 

/assign @JoshVanL

PR #456, opened 2018-08-21
Expose scale down utilization threshold

### LINKED ISSUE #340: Expose --scale-down-utilization-threshold cluster-autoscaler flag

/kind feature

**Description**: Expose --scale-down-utilization-threshold cluster-autoscaler flag in `tarmak.yaml`

JoshVanL:
Exposes scale down utilization threshold on cluster-autoscaler in the .tamak.yaml

fixes #340 

```release-note
Exposes auto-cluster's --scale-down-utilization-threshold in .tarmak.yaml
```

/assign @dippynark 
/cc @dippynark


pkg/puppet/puppet.go
```diff
@@ -170,6 +170,9 @@ func kubernetesClusterConfigPerRole(conf *clusterv1alpha1.ClusterKubernetes, rol
 		if conf.ClusterAutoscaler.Version != "" {
 			hieraData.variables = append(hieraData.variables, fmt.Sprintf(`kubernetes_addons::cluster_autoscaler::version: "%s"`, conf.ClusterAutoscaler.Version))
 		}
+		if conf.ClusterAutoscaler.ScaleDownThreshold != 0 {
```
dippynark: we should check here that the threshold is not less than 0 or greater than 1, or put something in https://github.com/jetstack/tarmak/blob/master/pkg/tarmak/cluster/cluster.go to error when that isn't the case

puppet/modules/kubernetes_addons/manifests/cluster_autoscaler.pp
```diff
@@ -11,6 +11,7 @@
   Optional[Boolean] $enable_overprovisioning=undef,
   Optional[String] $proportional_image=undef,
   Optional[String] $proportional_version=undef,
+  $scale_down_utilization_threshold = undef,
```
dippynark: Add type

puppet/modules/kubernetes_addons/templates/cluster-autoscaler-deployment.yaml.erb
```diff
@@ -73,9 +73,12 @@ spec:
 <% if @_enable_overprovisioning -%>
         - --expendable-pods-priority-cutoff=-10
 <% end -%>
+<% if @_enable_scale_down_utilization_threshold -%>
```
MattiasGees: I think you can just do `<% if @scale_down_utilization_threshold -%>`, there is no need for the extra variable. `undef` will return false while any other value will return true.

pkg/puppet/puppet.go
```diff
@@ -170,6 +170,9 @@ func kubernetesClusterConfigPerRole(conf *clusterv1alpha1.ClusterKubernetes, rol
 		if conf.ClusterAutoscaler.Version != "" {
 			hieraData.variables = append(hieraData.variables, fmt.Sprintf(`kubernetes_addons::cluster_autoscaler::version: "%s"`, conf.ClusterAutoscaler.Version))
 		}
+
+		hieraData.variables = append(hieraData.variables, fmt.Sprintf(`kubernetes_addons::cluster_autoscaler::scale_down_utilization_threshold: %v`, *conf.ClusterAutoscaler.ScaleDownThreshold))
```
dippynark: I think a check is needed here in case ScaleDownThreshold is nil?

dippynark:
@JoshVanL  looks good, just Mattias's suggestion and also a check for nil (I think)

pkg/puppet/puppet.go
JoshVanL: Shouldn't need a check since it's set in defaults but couldn't hurt

dippynark: Ah yeah, good point

PR #461, opened 2018-08-23
Mount NVMe drivers with docker through puppet

### LINKED ISSUE #449: Data volume doesn't mount correctly with NVMe drives

>
/kind bug
> /kind feature

**What happened**:
The extra data EBS volume drive isn't mapped correct to the devicemapper of Docker. This only happens on NVMe instances (C5d or M5d).
It uses the NVMe drive instead.

```
Filesystem Size Used Avail Use% Mounted on
/dev/nvme0n1p1 100G 11G 90G 11% /
...
overlay 100G 11G 90G 11% 
/var/lib/docker/overlay2/c0e310f522c92f55551841e1df6e209dd9cb0a14c9b70f31a0647cdccd6d27e8/merged
overlay 100G 11G 90G 11% /var/lib/docker/overlay2/4ea68503df8e17e5545ef98de154215095c05ce9dde80588407288ddc4c46e1a/merged
...
/dev/nvme2n1 296G 4.1G 277G 2% /var/lib/kubelet/plugins/kubernetes.io/aws-ebs/mounts/aws/eu-west-2b/vol-xxxxxxxxxxx
```

Docker uses EBS volume and not the NVMe drive,

Bring up a cluster with worker instances of C5d.large and start some pods. 

This is possibly also happening on other types where we use custom volumes.

**Environment**:
- Install tools:
- Others:

JoshVanL:
Mount NVMe drives to docker correctly. Only uses puppet to determine target drive.

fixes #449 

```release-note
Correctly mount docker storage on NVMe driver AWS instances.
```
/assign @simonswine


puppet/modules/aws_ebs/functions/disks.pp
```diff
@@ -0,0 +1,7 @@
+function aws_ebs::disks() >> Array[String] {
+  if $::disks == undef {
+    []
+  } else {
+    $::disks.keys.sort
```
simonswine: This will fail for more than 10 nvme disks

```
echo nvme0n1 nvme1n1 nvme2n1 [...] nvme10n1 nvme11n1 | xargs -n 1 | sort
nvme0n1
nvme10n1
nvme11n1
nvme1n1
nvme2n1
[...]
```

This should be the correct order:

```
nvme0n1
nvme1n1
nvme2n1
[...]
nvme10n1
nvme11n1
```

We should address and test this

simonswine:
Thanks this works exactly as expected!

/approve
/lgtm

PR #462, opened 2018-08-23
Fix code generation

simonswine:
```release-note
Ensure code generation is verified correctly 
```


JoshVanL:
vet fails on go 1.10, fine for 1.9.
/assign @simonswine 
/unassign

PR #473, opened 2018-08-29
Consul puppet module

JoshVanL:
Adds a standalone consul module to puppet

```release-note
NONE
```


puppet/modules/consul/.fixtures.yaml
```diff
@@ -0,0 +1,7 @@
+fixtures:
+  symlinks:
+    consul: "#{source_dir}"
+    archive: "#{source_dir}/../archive"
+    stdlib: "#{source_dir}/../stdlib"
+    airworthy: "#{source_dir}/../airworthy"
+    aws_ebs: "#{source_dir}/../aws_ebs"
```
simonswine: remove that module (see below)

puppet/modules/consul/templates/consul.json.epp
```diff
@@ -0,0 +1,7 @@
+<%- | $config | -%>
```
simonswine: can we use an erb file instead and use `<%= require "json"; JSON.pretty_generate(@config) %>`

IMHO this will do the sorting for us as well

puppet/modules/consul/templates/consul.service.erb
```diff
@@ -0,0 +1,19 @@
+[Unit]
```
simonswine: The whole file needs to be parametrized

puppet/modules/consul/templates/download-vault-consul.sh.erb
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
```
simonswine: remove that file

puppet/modules/consul/templates/ensure-ebs-volume-formatted.service.erb
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
simonswine: remove that file

puppet/modules/consul/templates/var-lib-consul.mount.erb
```diff
@@ -0,0 +1,9 @@
+[Unit]
```
simonswine: remove that file

puppet/modules/consul/files/consul-backup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
simonswine: The double `$$` are from terraform escaping, we should at least fix those

puppet/modules/consul/templates/consul.service.erb
simonswine: https://github.com/jetstack/tarmak/blob/71e9b872b3d98adc0b4b08f428192020129fb4bb/puppet/modules/consul/templates/consul.service.erb

puppet/modules/consul/manifests/init.pp
```diff
@@ -0,0 +1,106 @@
...
+    gid    => $gid,
+  }
+  -> user { $user:
+    ensure => present,
+    uid    => $uid,
+    shell  => $nologin,
+    home   => $data_dir,
+  }
+
+  class { '::airworthy': }
```
simonswine: Can finish off the init.pp like that:

```
  # install airworthy if necessary
  ensure_resource('class', '::airworthy', {})

  Class['::airworthy']
  -> class { '::consul::install': }
  -> class { '::consul::config': }
  ~> class { '::consul::service': }
  -> Class['::consul']
```

Otherwise you can only have exactly a single module using airworthy

simonswine: Upper case Class uses a reference (that needs to exist before), while class is loading the classin

puppet/modules/consul/spec/classes/config_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    it { should contain_class('consul::config') }
+
+    it 'should install consul config' do
+      should contain_file('/etc/consul').with(
+        :ensure => 'directory',
+      )
+      should contain_file('/etc/consul/consul.json').with(
+        :mode => '0600',
+      )
+      should contain_file('/etc/consul/consul.json')
```
simonswine: Get the content of the file and parse it as json and then make sure the content is correct.

We do that with YAMLs here:
https://github.com/jetstack/tarmak/blob/master/puppet/modules/kubernetes_addons/spec/classes/grafana_spec.rb

puppet/modules/consul/.fixtures.yaml
```diff
@@ -0,0 +1,7 @@
+fixtures:
```
simonswine: The whole file should be called .fixtures.yml and you should have the exact same file in `.fixtures.yml.local`

puppet/modules/consul/manifests/service.pp
```diff
@@ -0,0 +1,107 @@
...
+  }
+  ~> service { "${service_name}.service":
+    ensure  => running,
+    enable  => true,
+    require => [
+      Exec["${service_name}-systemctl-daemon-reload"],
+      ],
+  }
+
+  $disks = aws_ebs::disks()
```
simonswine: this should happen in class `tarmak::vault` and only for AWS provider

simonswine:
The three nodes acceptance tests somehow is not building a proper cluster (or I don't know how to auth against it:

Somehow I think the consul cluster of the three nodes is not properly up

```$ bundle exec rake beaker:ssh:3n-cluster[vault-1]                                                                                                                                                                                                                                                                    1 ↵
Last login: Thu Aug 30 15:18:21 2018 from 192.168.121.1
[vagrant@vault-1 ~]$ sudo -i
[root@vault-1 ~]# cat /etc/consul/consul.json
{
  "acl_default_policy": "deny",
  "acl_down_policy": "deny",
  "acl_master_token": "",
  "acl_agent_token": "",
  "acl_datacenter": "dc1",
  "datacenter": "dc1",
  "log_level": "INFO",
  "client_addr": "127.0.0.1",
  "bind_addr": "0.0.0.0",
  "advertise_addr": "10.123.0.11",
  "encrypt": "",
  "bootstrap_expect": 3,
  "server": true,
  "disable_remote_exec": true,
  "retry_join": [
    "10.123.0.11",
    "10.123.0.12",
    "10.123.0.13"
  ],
  "ca_file": "/etc/consul/consul-ca.pem",
  "cert_file": "/etc/consul/consul.pem",
  "key_file": "/etc/consul/consul-key.pem",
  "verify_outgoing": true,
  "verify_incoming": true
}
[root@vault-1 ~]# consul member list```

the member list is empty and also the token is not picked up correctly
The test itself is fine, but it's not really having a check if consul is up and ready

/assign @JoshVanL 
/unassign

PR #476, opened 2018-08-31
vault_server puppet module

JoshVanL:
Creates a vault_server puppet module to be later used instead of cloud-init to install vault servers.

```release-note
Add vault_server puppet module to initiate vault servers
```


puppet/modules/consul/manifests/init.pp
```diff
@@ -120,7 +120,9 @@
   }
 
   # install airworthy if necessary
-  ensure_resource('class', '::airworthy', {})
+  if !defined(Class[::airworthy]) {
+    ensure_resource('class', '::airworthy', {})
+  }
```
simonswine: Where is that change coming from?

I prefer the old variant. It feels like doing the if not defined step twice

puppet/modules/vault_server/.fixtures.yml
```diff
@@ -0,0 +1,7 @@
+fixtures:
+  symlinks:
+    vault_server: "#{source_dir}"
+    archive: "#{source_dir}/../archive"
+    stdlib: "#{source_dir}/../stdlib"
+    airworthy: "#{source_dir}/../airworthy"
+    tarmak: "#{source_dir}/../tarmak"
```
simonswine: vault_server should be usable without importing tarmak module

puppet/modules/vault_server/.fixtures.yml.local
```diff
@@ -0,0 +1,7 @@
+fixtures:
+  symlinks:
+    vault_server: "#{source_dir}"
+    archive: "#{source_dir}/../archive"
+    stdlib: "#{source_dir}/../stdlib"
+    airworthy: "#{source_dir}/../airworthy"
+    tarmak: "#{source_dir}/../tarmak"
```
simonswine: vault_server should be usable without importing tarmak module

puppet/modules/vault_server/spec/classes/init_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    it 'should create vault user' do
+      should contain_user(app_name).with(
+        :home => lib_dir,
+      )
+    end
+
+    it 'should mount vault ebs device' do
+      should contain_file('/etc/systemd/system/attach-ebs-volume-vault.service')
+      should contain_file('/etc/systemd/system/ensure-ebs-volume-vault-formatted.service')
```
simonswine: no ebs attachment in vault_server

puppet/modules/vault_server/manifests/init.pp
```diff
@@ -0,0 +1,66 @@
...
+  user { $app_name:
+    ensure => 'present',
+    system => true,
+    home   => $lib_dir,
+  }
+
+  # install airworthy if necessary
+  if !defined(Class[::airworthy]) {
+    ensure_resource('class', '::airworthy', {})
+  }
```
simonswine: See comments from the init.pp of consul

puppet/modules/vault_server/manifests/init.pp
```diff
@@ -0,0 +1,66 @@
...
+
+  Class['::airworthy']
+  ~> class {'::tarmak::vault':
```
simonswine: Vault server should work without tarmak

PR #486, opened 2018-09-05
Applies hot fix for class loop in airworthy module

JoshVanL:
Airworthy puppet module had a class loop which this PR fixes

```release-note
NONE
```

/assign @dippynark @simonswine

### DIFF
--- a/puppet/modules/airworthy/manifests/init.pp
+++ b/puppet/modules/airworthy/manifests/init.pp
@@ -30,4 +30,5 @@ class airworthy (
   }
 
   class { '::airworthy::install': }
+  -> Class['::airworthy']
 }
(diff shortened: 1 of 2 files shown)


puppet/modules/airworthy/manifests/init.pp
```diff
@@ -30,5 +30,4 @@
   }
 
   class { '::airworthy::install': }
-  -> Class['::airworthy']
```
simonswine: Can we please keep that? Otherwise the dependencies will be messed up and

dippynark:
/assign @JoshVanL 
some build errors which look transient

simonswine:
Thanks @JoshVanL + @dippynark

PR #493, opened 2018-09-06
Only converge on wing when not in infrastructure only

### LINKED ISSUE #475: Don't wait for wing to converge when running apply with --infrastructure only
**What happened**:
I ran `tarmak clusters apply --infrastructure-only` on a cluster where wing was still running, my apply command waited for wing to complete.

I expected to not have to wait for wing. I'd viewed this just as a kind of terraform only flag and expected it to complete when terraform was done.

Use the `--infrastructure-only` flag on an apply operation while wing is still running.

**Environment**: 
- Kubernetes version (use `kubectl version`): 1.10
- Tarmak: 0.5 alpha 2

In `pkg/tarmak/terraform.go` we have some logic that looks a bit like this:

```go
func (t *Tarmak) CmdTerraformApply(args []string, ctx context.Context) error {
...
	t.cluster.Log().Info("running apply")
	// run terraform apply always, do not run it when in configuration only mode
	if !t.flags.Cluster.Apply.ConfigurationOnly {
		err := t.terraform.Apply(t.Cluster())
	}

	// upload tar gz only if terraform hasn't uploaded it yet
	if t.flags.Cluster.Apply.ConfigurationOnly {
		err := t.Cluster().UploadConfiguration()
	}

	// reapply config expect if we are in infrastructure only
	if !t.flags.Cluster.Apply.InfrastructureOnly {
		err := t.Cluster().ReapplyConfiguration()
	}

	// wait for convergance in every mode  <----------- Are we sure we want this?
	err := t.Cluster().WaitForConvergance()
}
```

I propose that when doing an InfrastructureOnly apply, we should skip waiting for convergence. Perhaps there's something I'm missing. What do people think?

JoshVanL:
Don't wait for wing to converge when we're not in infrastructure only

fixes #475

```release-note
Don't wait for wing conversion when we're not in infrastructure only
```

### DIFF
--- a/pkg/tarmak/cmd.go
+++ b/pkg/tarmak/cmd.go
@@ -53,7 +53,8 @@ func (c *CmdTarmak) Apply() error {
 		return err
 	}
 
-	hasChanged := false
+	// assume a change so that we wait for convergence in configuration only
+	hasChanged := true
 	// run terraform apply always, do not run it when in configuration only mode
 	if !c.flags.Cluster.Apply.ConfigurationOnly {
 		hasChanged, err = c.terraform.Apply(c.Cluster())
--- a/pkg/tarmak/cmd.go
+++ b/pkg/tarmak/cmd.go
@@ -82,10 +82,12 @@ func (c *CmdTarmak) Apply() error {
 	default:
 	}
 
-	// wait for convergance in every mode
-	err := c.Cluster().WaitForConvergance()
-	if err != nil {
-		return err
+	// wait for convergance if not in infrastructure only
+	if !c.flags.Cluster.Apply.InfrastructureOnly {
+		err := c.cluster.WaitForConvergance()
+		if err != nil {
+			return err
+		}
 	}
 
 	return nil
--- a/pkg/tarmak/cmd.go
+++ b/pkg/tarmak/cmd.go
@@ -48,13 +48,15 @@ func (c *CmdTarmak) Plan() (returnCode int, err error) {
 }
 
 func (c *CmdTarmak) Apply() error {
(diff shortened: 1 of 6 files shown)


simonswine:
Can you look at this together with our notes from the tarmak call last week:

- Add `--wait-for-convergence=true` flag that can be enabled
- If plan doesn't not change anything don't wait for convergance (this is blocked by #531)

MattiasGees:
Nothing changed, only rebase

/lgtm
/approve

pkg/tarmak/cmd.go
```diff
@@ -82,10 +84,12 @@ func (c *CmdTarmak) Apply() error {
 	default:
 	}
 
-	// wait for convergance in every mode
-	err := c.Cluster().WaitForConvergance()
-	if err != nil {
-		return err
+	// wait for convergance if flag enabled and has changed
+	if hasChanged && c.flags.Cluster.Apply.WaitForConvergence {
```
simonswine: We need to wait for convergance when we update the config through the tarmak update method  (and not terraform) in ConfigurationOnly 

We update the puppet tar config and send out a converge request. The only way to skip that would if we would know if the config has changed through our upload method in tarmak (puppet.tar.gz is the same).

When we update the puppet.tar.gz in the default mode we have a terraform change for updating the puppet.tar.gz

simonswine:
I don't think that handles `cluster apply -C` correctly. I think in those cases it needs to be waiting for convergence as the configuration might have changed. (see my code commit)

/assign @JoshVanL 
/unassign

simonswine:
Thank you that seems to work

/approve
/lgtm

PR #495, opened 2018-09-07
Add Terraform debug shell error when binary version is incompatible

### LINKED ISSUE #490: Tarmak should error when terraform debug shell is opened with local binary having wrong version

/kind feature

**What happened**: Using a higher version of terraform in the terraform debug shell than what is vendored into Tarmak causes the state version to be updated preventing Tarmak from running and further terraform commands

**What you expected to happen**: Error is the local versions is different

**How to reproduce it (as minimally and precisely as possible)**: Use terraform 0.11.8 with Tarmak version 0.5.0-alpha2

dippynark:
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #490 

```release-note
Add Terraform debug shell error when binary version is incompatible
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2075,6 +2075,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "31fcb889e654aad64956d35d9c816352d583e517646af9ad23ca16a77727d2d4"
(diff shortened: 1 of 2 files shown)


JoshVanL:
@dippynark Looks good, just found the rouge append.
/unassign
/assign @dippynark

simonswine:
Apart from josh's append line this looks good and works well for me

/approve

PR #498, opened 2018-09-07
Download backinator binary in consul puppet module

JoshVanL:
The consul backinator util tool is not installed within the consul module

```release-note
NONE
```


puppet/modules/consul/spec/classes/install_spec.rb
```diff
@@ -43,5 +43,16 @@
         :target => '/opt/consul-1.2.1/consul-backup.sh',
       )
     end
+
+    it 'should install consul backinator' do
+      should contain_file('/opt/consul-1.2.1/consul-backinator').with(
```
simonswine: I think if we don't have the version in the path we might get into trouble updating consul-backinator. We also download a new backinator when we updating consul. I suggest we do rather something:
- `/opt/consul-backinator-0.3.0/consul-backinator`

(Also check the paths for consul-exporter)

simonswine:
/approve
/lgtm

Thanks @JoshVanL

PR #501, opened 2018-09-10
Change threshold to greater than or equal to

dippynark:
**What this PR does / why we need it**: The `StatusCheckFailed_System` is a binary value so we must check for when the metric equals one

```release-note
Fix autorecovery threshold for cloudwatch alarms
```


simonswine:
/assign @dippynark 

A bit more background: how have you discovered this?. 

The UI shows it somehow like that: `StatusCheckFailed_System > 1 for 2 datapoints within 2 minutes`

I might have understood it. The sum of 2 datapoints needs to be `> 1`

dippynark:
@simonswine these blog posts https://aws.amazon.com/blogs/aws/new-auto-recovery-for-amazon-ec2/ and https://aws.amazon.com/blogs/aws/ec2-instance-status-metrics/

Not sure where you are seeing that but it may be grouping them above certain periods. The metric we are setting up is using the minimum so it'd give different results

PR #502, opened 2018-09-10
Restart Consul on failure

dippynark:
**What this PR does / why we need it**: If Consul fails we want to restart it

```release-note
Restart Consul on failure
```


simonswine:
/approve

/assign @dippynark 
/unassign

@dippynark Can you also make the change here:
https://github.com/jetstack/tarmak/blob/master/puppet/modules/consul/templates/consul.service.erb

PR #503, opened 2018-09-10
Upgrade calico to 3.1.3

simonswine:
```release-note
Update calico version to 3.1.3
Change tolerations of calico-node to support NoExecute taints in 1.12+
```


simonswine:
/assign @JoshVanL 
/unassign 

Can you give that a look/try. The additional toleration is required for 1.12+

PR #510, opened 2018-09-11
Restart bastion etcd and wing-server on failure

dippynark:
Signed-off-by: Luke Addison <luke.addison@jetstack.io>

**What this PR does / why we need it**: Restarts etcd and wing-server when they fail

```release-note
Restart etcd and wing-server on the bastion automatically on failure
```


simonswine:
Thanks @dippynark

PR #513, opened 2018-09-12
Implement Vault runcmd commands as systemd units

### LINKED ISSUE #506: Implement AMI Vault user data runcmd commands as oneshot systemd units

/kind feature

**What happened**: My Vault cluster didn't start up because one of the [`runcmd`](https://github.com/jetstack/tarmak/blob/master/terraform/amazon/modules/vault/templates/vault_user_data.yaml#L378) commands failed to complete (reason unknown)

**What you expected to happen**: `runcmd` does not retry if commands fail so we should move that responsibility to systemd and run those scripts as a oneshot unit that will restart if there is a failure. This will allow us to only run `systemctl enable` and `systemctl start` commands that do not depend on the network.

dippynark:
Signed-off-by: Luke Addison <luke.addison@jetstack.io>

**What this PR does / why we need it**: Intermittent issues (e.g. network) can cause Vault's runcmd userdata commands to fail. This PR fixes this by automatically restarting them if they fail

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #506 

**Release note**:
```release-note
Run Vault userdata runcmd commands as systemd units
```

### DIFF
--- a/terraform/amazon/modules/vault/templates/vault_user_data.yaml
+++ b/terraform/amazon/modules/vault/templates/vault_user_data.yaml
@@ -367,6 +367,57 @@ write_files:
     export PATH=$PATH:/usr/local/bin
     export VAULT_ADDR=https://localhost:8200
     export VAULT_CACERT=/etc/vault/tls/ca.pem
+- path: /etc/systemd/system/download-vault-consul.service
+  permissions: '0644'
+  content: |
+    [Unit]
+    Description=Download Vault and Consul
+    After=network-online.target
+    Requires=network-online.target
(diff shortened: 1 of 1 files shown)


dippynark:
@simonswine if this is something we need, I can do the same for the rest of the `runcmd` part (e.g. yum installs)

simonswine:
Will be not merged, because it creates some risks and 0.6.0 will have that part in puppet anyhow

PR #522, opened 2018-09-14
Adds force-unlock terraform command

### LINKED ISSUE #509: Add unlock command

/kind feature

**What happened**: It is possible to not release the Dynamo lock when tarmak crashes or is force stopped

**What you expected to happen**: We should add a `tarmak clusters unlock` command to release the lock manually. We should include a warning to describe the possible dangers of doing this

JoshVanL:
Adds a tarmak cluster force-unlock to perform a terraform force-unlock on it's remote state. If no lock ID is given then it will retrieve the remote lock ID from DynamoDB. Tarmak will ask if sure before applying.

fixes #509 

```release-note
Adds cluster force-unlock subcommand for to release terraform state lock.
```

### DIFF
--- a/cmd/tarmak/cmd/cluster_force-unlock.go
+++ b/cmd/tarmak/cmd/cluster_force-unlock.go
@@ -1,0 +1,25 @@
+// Copyright Jetstack Ltd. See LICENSE for details.
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/jetstack/tarmak/pkg/tarmak"
+)
+
+var clusterForceUnlockCmd = &cobra.Command{
+	Use:   "force-unlock",
+	Short: "Remove remote lock using lock ID",
+	Run: func(cmd *cobra.Command, args []string) {
+		t := tarmak.New(globalFlags)
+		defer t.Cleanup()
+
+		forceUnlockCmd := t.NewCmdTerraform(args)
+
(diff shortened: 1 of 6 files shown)


simonswine:
@JoshVanL thanks for adding that feature

This looks good to me. Can we have some docs for that?

simonswine:
Thinking a bit more about it, why do we not just use the terraform implementation and pass that one throug?

JoshVanL:
@simonswine I've been looking though the terraform code and the problem with looking up the lock ID that way is that it is very involved to get terraform initialized with the backend. A lot of the logic is in top level main meaning it'll have to be straight copied over - a problem that keeps coming up... I don't know of a way to get the lock ID out of a cli command without doing a horrible string search on the CLI output on failure. Seems like a lot of extra code + complexity for terraform to basically just be doing the same dynamo lookup, WDYT?

/unassign
/assign @simonswine

simonswine:
I think terraform has valid reasons why it's not auto unlocking by a single command and expecting the user to actually copy paste the lock id them selves. I think we should have subcommand that behvaes exactly like `terraform force-unlock`. If someone is already using terraform he is used to that behaviour and we don't have the risk to accidentally removing a lock

simonswine:
Thank you @JoshVanL , lgtm!

/approve
/lgtm

PR #525, opened 2018-09-17
Adds hot fix to remote state variable order

JoshVanL:
Fixes variable order in RemoteStateName fmt.Sprintf

```release-note
NONE
```
/assign @simonswine


simonswine:
Thanks for breaking and fixing it :joy: @JoshVanL 

/approve
/lgtm

PR #526, opened 2018-09-17
Ensure validate/verify is run during a debug shell

JoshVanL:
Validate / Verify was not being run on a `cluster debug terraform shell` meaning that the remote KMS for the lock was never being retrieved.

```release-note
NONE
```


simonswine:
Thanks for that clean-up

/approve 
/lgtm

Issue #523, opened 2018-09-14
Tarmak E2E tests on AWS


simonswine:
We need to ensure that tarmak and various high level deployment options are tested continously against the AWS cloud provider. The main tests scenarios are:

- [ ] SingleClusterEnvironment-SingleZone
- [ ]  MultiClustersEnvironment-MultiZones
- [ ] Upgrade from the latest minor version of tarmak

simonswine:
I have started with the implementation of those tests. See #518. I think the main point they are missing currently is an ability to clean up all resources that tarmak creates. (terraform state S3 bucket and DynamoDB table. SSH keys uploaded.

I think we should have a built in command in tarmak to delete an environment. (All of those resources are bound to an environment). Which will clear those resources and also remove associated clusters from tarmak.yaml

PR #527, opened 2018-09-17
Clean-up when destroying environment

MattiasGees:
This will destroy a complete environment and all of its clusters:
- Run terraform destroy
- Remove state files from S3 and dynamodb
- Move environment folder to .archive
- Remove environment config from tarmak.yaml

part of #523 

**Release note**:

```release-note
Add terraform environment destroy command
```


pkg/tarmak/terraform.go
```diff
@@ -179,6 +179,13 @@ func (c *CmdTerraform) Destroy() error {
 		return err
 	}
 
+	c.tarmak.cluster.Log().Info("Removing environment")
+
+	err = c.tarmak.environment.Remove()
```
MattiasGees: Not sure if we have to make this command part of `tarmak cluster destroy`. Input is welcome for this one!

MattiasGees: ok

MattiasGees:
Talked about this with the team. Need to make a few changes.
Change `--force` to `--auto-approve=false`. Together with this, we can use user interaction to run the destroy.

MattiasGees:
All functionality works. UI interaction is implemented. 

One task is remaining. When de skip the cluster destroy part, we need to make sure nothing is running on AWS anymore. Only after that we can delete the the config from tarmak.yaml and move the environment folder to `.archive`

MattiasGees:
This is ready for review
/unassign
/assign @simonswine

pkg/terraform/terraform.go
```diff
@@ -21,7 +22,6 @@ import (
 
 	"github.com/jetstack/tarmak/pkg/tarmak/interfaces"
 	"github.com/jetstack/tarmak/pkg/tarmak/utils"
-	"github.com/jetstack/tarmak/pkg/terraform/providers/tarmak/rpc"
```
simonswine: can we keep it like that?

cmd/tarmak/cmd/environment.go
```diff
@@ -11,6 +12,25 @@ var environmentCmd = &cobra.Command{
...
+		"auto-approve destroy of a complete environment",
+	)
+
+	fs.StringVarP(
+		&store.Name,
+		"name",
+		"n",
+		"",
+		"name of the environment to destroy",
+	)
```
simonswine: I would prefer to have that as an argument rather than a flag

This aligns closer with `clusters set-current test` and other commands

JoshVanL: I agree

pkg/tarmak/config/config.go
```diff
@@ -281,6 +281,30 @@ func (c *Config) AppendEnvironment(env *tarmakv1alpha1.Environment) error {
...
 }
 
+func (c *Config) RemoveEnvironment(environment string) error {
+	for key, env := range c.conf.Environments {
+		if env.Name == environment {
+			c.conf.Environments = append(c.conf.Environments[:key], c.conf.Environments[key+1:]...)
+		}
+	}
+
+	done := false
```
JoshVanL: This could be a bit nicer, something like 
```
	for i := 0; i < len(c.conf.Clusters); i++ {
		if c.conf.Clusters[i].Environment == environment {
			c.conf.Clusters = append(c.conf.Clusters[:i], c.conf.Clusters[i+1:]...)
			i -= 1
		}
	}
```

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,153 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+
+	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
+		Bucket: aws.String(a.RemoteStateName()),
+	})
+
+	if err != nil {
+		ec2err, ok := err.(awserr.Error)
+		if ok && ec2err.Code() == "BucketNotEmpty" {
+			resp, err := svc.ListObjectVersions(
+				&s3.ListObjectVersionsInput{
```
JoshVanL: Maybe this could go into a separate function, lots of tabs :grimacing:

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,153 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+	svc, err := a.DynamoDB()
+	if err != nil {
+		return err
+	}
+
+	_, err = svc.DeleteItem(&dynamodb.DeleteItemInput{
+		TableName: aws.String(a.RemoteStateName()),
+		Key: map[string]*dynamodb.AttributeValue{
+			"LockID": {
+				S: aws.String(fmt.Sprintf("%s/%s/%s/%s.tfstate-md5", a.RemoteStateName(), a.tarmak.Environment().Name(), a.tarmak.Cluster().Name(), "main")),
```
JoshVanL: Maybe make this Sprintf into a function

pkg/terraform/terraform.go
```diff
@@ -5,6 +5,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"github.com/jetstack/tarmak/pkg/terraform/providers/tarmak/rpc"
```
JoshVanL: :sob:

pkg/tarmak/tarmak.go
```diff
@@ -365,3 +374,135 @@ func (t *Tarmak) CmdKubectl(args []string) error {
 func (t *Tarmak) CancellationContext() interfaces.CancellationContext {
 	return t.ctx
 }
+
+func (t *Tarmak) DestroyEnvironment() error {
+	inputDestroy := input.New(os.Stdin, os.Stdout)
+
+	destroyClusters := true
```
JoshVanL: Perhaps create a `var err error` here so we don't have to create lots of response bool variables, i.e
```
var err error
destroy := true
...
destroy, err = inputDestroy(....
...
if destroy {...
.....
```

pkg/tarmak/tarmak.go
```diff
@@ -365,3 +374,135 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+		for _, cluster := range t.Environment().Clusters() {
+			hosts, err := cluster.ListHosts()
+			if err != nil {
+				return err
+			}
+			if len(hosts) > 0 {
+				return fmt.Errorf("can't proceed with destroying this environment, because it still has hosts running")
+			}
+		}
+	}
```
JoshVanL: Though I like that it is asking at every step if the user wishes to proceed, I would expect that a no answer would immediately return

pkg/tarmak/tarmak.go
```diff
@@ -365,3 +374,135 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+		removeConfig = r
+	}
+
+	if removeConfig {
+		t.log.Infof("Removing environment %v from tarmak.yaml", t.Environment().Name())
+
+		if err := t.config.RemoveEnvironment(t.Environment().Name()); err != nil {
+			return err
+		}
+	}
```
JoshVanL: The current cluster and environment need to be unset or set to something else

pkg/tarmak/tarmak.go
```diff
@@ -365,3 +374,135 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+			Query:   "Remove environment from tarmak.yaml?",
+		})
+		if err != nil {
+			return err
+		}
+		removeConfig = r
+	}
+
+	if removeConfig {
+		t.log.Infof("Removing environment %v from tarmak.yaml", t.Environment().Name())
```
JoshVanL: Maybe use all lower case on all 'infof's to match what we are doing elsewhere

pkg/tarmak/tarmak.go
```diff
@@ -365,3 +374,135 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+	inputDestroy := input.New(os.Stdin, os.Stdout)
+
+	destroyClusters := true
+	moveFolder := true
+	removeConfig := true
+
+	if !t.flags.Environment.Destroy.AutoApprove {
+		d, err := inputDestroy.AskYesNo(&input.AskYesNo{
+			Default: false,
+			Query:   "Destroy all clusters?",
```
JoshVanL: Maybe print out the environment name again here to triple check the user `Destroy all clusters in environment env?`

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+
+	result, err := svc.Scan(&dynamodb.ScanInput{
+		TableName: aws.String(a.RemoteStateName()),
+	})
+
+	if err != nil {
+		return 1, err
+	}
+
+	return *result.Count, nil
```
JoshVanL: Is this ever nil?

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+	if err != nil {
+		ec2err, ok := err.(awserr.Error)
+		if ok && ec2err.Code() == "BucketNotEmpty" {
+			resp, err := a.listObjectVersionsRemoteState()
+
+			if err != nil {
+				return fmt.Errorf("error listing objects to delete: %s", err)
+			}
+
+			objectsToDelete := a.objectsToDeleteFromRemoteState(resp)
```
JoshVanL: variable needed here?

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+
+			params := &s3.DeleteObjectsInput{
+				Bucket: aws.String(a.RemoteStateName()),
+				Delete: &s3.Delete{
+					Objects: objectsToDelete,
+				},
+			}
+
+			_, err = svc.DeleteObjects(params)
+
```
JoshVanL: newline?

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
+		Bucket: aws.String(a.RemoteStateName()),
+	})
+
+	if err != nil {
+		ec2err, ok := err.(awserr.Error)
+		if ok && ec2err.Code() == "BucketNotEmpty" {
+			resp, err := a.listObjectVersionsRemoteState()
+
```
JoshVanL: new line?

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+	}
+
+	return nil
+
+}
+
+func (t *Tarmak) DestroyActivecluster() error {
+	var args []string
+
+	t.log.Infof("Destroying cluster %v", t.Cluster().Name())
```
JoshVanL: Some upper cases still

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+		if err = t.config.RemoveEnvironment(t.Environment().Name()); err != nil {
+			return err
+		}
+	}
+
+	return nil
+
+}
+
+func (t *Tarmak) DestroyActivecluster() error {
```
JoshVanL: DestroyActivecuster -> DestroyActiveCluster

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+	var args []string
```
JoshVanL: variable needed?

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+
+	t.log.Infof("Destroying cluster %v", t.Cluster().Name())
+	destroyCmd := t.NewCmdTerraform(args)
```
JoshVanL: We are not passing signals properly here. The destroy was not respecting my interrupts

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+	if err := destroyCmd.Destroy(); err != nil {
+		return err
+	}
+
+	t.log.Infof("Removing S3 state and dynamoDB lock for %v", t.Cluster().Name())
+	if err := t.environment.Provider().Remove(); err != nil {
+		return err
+	}
```
JoshVanL: We should probably be checking context here for signals

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
 	return t.ctx
 }
+
+func (t *Tarmak) DestroyEnvironment() error {
+	inputDestroy := input.New(os.Stdin, os.Stdout)
+
+	var err error
+
+	destroyClusters := true
+	moveFolder := true
```
JoshVanL: We can use one response variable here

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+		archivePath := filepath.Join(t.ConfigPath(), ".archive")
+		environmentArchivePath := filepath.Join(archivePath, t.Environment().Name())
+
+		if _, err = os.Stat(archivePath); err != nil {
+			if os.IsNotExist(err) {
+				if err = os.MkdirAll(archivePath, os.ModePerm); err != nil {
+					return err
+				}
+			} else {
+				return fmt.Errorf("uncatched error: %v", err)
```
JoshVanL: uncatched?

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	if moveFolder {
+		t.log.Info("moving environment folder to .archive")
+
+		archivePath := filepath.Join(t.ConfigPath(), ".archive")
```
JoshVanL: The path would be nice to set as a flag

pkg/tarmak/tarmak.go
```diff
@@ -357,3 +366,134 @@ func (t *Tarmak) CmdKubectl(args []string) error {
...
+				}
+			}
+		}
+
+		// After destroying all other clusters, we can destroy the hub
+		t.cluster = t.Environment().Hub()
+		if err = t.DestroyActivecluster(); err != nil {
+			return err
+		}
+	} else {
```
JoshVanL: I still think that a no here should immediately exit

pkg/tarmak/tarmak.go
```diff
@@ -141,24 +142,33 @@ func (t *Tarmak) initializeModules() {
...
+	t.environment, err = t.EnvironmentByName(currentEnvironmentName)
 	if err != nil {
-		return fmt.Errorf("failed to retrieve current cluster name: %s", err)
+		return err
 	}
-	// init cluster
-	t.cluster, err = t.environment.Cluster(clusterName)
-	if err != nil {
-		return fmt.Errorf("error finding current cluster '%s': %s", clusterName, err)
+	if !environmentDestroy {
```
JoshVanL: Why is this needed?

pkg/tarmak/config/config.go
```diff
@@ -281,6 +281,38 @@ func (c *Config) AppendEnvironment(env *tarmakv1alpha1.Environment) error {
...
+	}
+
+	// Set currentCluster to another cluster if it is part of the environment we want to delete
+	currentConfigEnvironment, err := c.CurrentEnvironmentName()
+	if err != nil {
+		return err
+	}
+	// Pick the first found environment/cluster to replace currentCluster
+	if currentConfigEnvironment == environment {
+		for _, cluster := range c.conf.Clusters {
```
JoshVanL: Why can't we just use `	c.SetCurrentCluster(c.tarmak.Clusters()[0].ClusterName())` ?

pkg/tarmak/provider/amazon/tf_remote_state.go
MattiasGees: an `int64` can't be nil.

pkg/tarmak/tarmak.go
MattiasGees: Because we give an environment as input, we don't really set an cluster. It tried to pick the cluster from the config, which can fail if the selected cluster is not part of the environment was have as user input.

Going to think a bit more about it on how to solve this in a nicer way. Suggestions are welcome.

pkg/tarmak/tarmak.go
MattiasGees: Maybe `unknown` is better?

pkg/tarmak/tarmak.go
MattiasGees: It already kinda is because we use the configpath. I still like to keep all config together even when it is archived.

pkg/tarmak/provider/amazon/tf_remote_state.go
JoshVanL: result.Count is a pointer to an int64 which you are dereferencing

pkg/tarmak/tarmak.go
JoshVanL: What's wring with 'failed to get status of file: %s"?

pkg/tarmak/tarmak.go
JoshVanL: Okay. I think we should have some backup naming scheme if the file is already existing then

pkg/tarmak/tarmak.go
JoshVanL: I see, that makes sense. Will have a think

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+		for _, v := range resp.DeleteMarkers {
+			objectsToDelete = append(objectsToDelete, &s3.ObjectIdentifier{
+				Key:       v.Key,
+				VersionId: v.VersionId,
+			})
+		}
+	}
+
+	if len(resp.Versions) != 0 {
```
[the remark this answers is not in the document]
MattiasGees: When you delete a object, we also have to delete all versions of it before we can remove a bucket.

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,175 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+		return nil, err
+	}
+	return resp, nil
+}
+
+// Create a list of objects to delete based on a S3 list object
+func (a *Amazon) objectsToDeleteFromRemoteState(resp *s3.ListObjectVersionsOutput) []*s3.ObjectIdentifier {
+	objectsToDelete := make([]*s3.ObjectIdentifier, 0)
+
+	if len(resp.DeleteMarkers) != 0 {
```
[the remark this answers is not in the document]
MattiasGees: Every deleted object in a versioned bucket has a deleteMarker as it is not fully deleted. In the docs to delete a bucket, it says that we need to be sure that all are gone. https://docs.aws.amazon.com/sdk-for-go/api/service/s3/#S3.DeleteBucket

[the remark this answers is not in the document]
JoshVanL: Sure but checking the length of a slice before a range is totally unnecessary

pkg/tarmak/provider/amazon/tf_remote_state.go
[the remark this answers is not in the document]
JoshVanL: Also please remove

pkg/tarmak/tarmak.go
MattiasGees: Going to use this solution https://github.com/jetstack/tarmak/pull/575#discussion_r225991922

cmd/tarmak/cmd/environment_destroy.go
```diff
@@ -0,0 +1,35 @@
...
+
+// environmentDestroyCmd handles `tarmak environment destroy`
+var environmentDestroyCmd = &cobra.Command{
+	Use:   "destroy [NAME]",
+	Short: "Destroy a environment",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
+			return errors.New("you have to give the environment name")
+		}
+		globalFlags.Environment.Destroy.Name = args[0]
```
JoshVanL: We shouldn't need to use global flags here, we pass through the args. We should also check == 1 instead of >0

pkg/tarmak/cmd.go
```diff
@@ -103,6 +104,118 @@ func (c *CmdTarmak) Destroy() error {
...
+		// After destroying all other clusters, we can destroy the hub
+		c.cluster = c.Environment().Hub()
+		c.log.Infof("destroying cluster %v", c.Cluster().Name())
+		if err := c.Destroy(); err != nil {
+			return err
+		}
+		if err = c.DestroyClusterMetadata(); err != nil {
+			return err
+		}
+	} else {
```
JoshVanL: I still think a no response to deleting all clusters should immediately exit after a no response

pkg/tarmak/cmd.go
```diff
@@ -103,6 +104,118 @@ func (c *CmdTarmak) Destroy() error {
...
+		response, err = inputDestroy.AskYesNo(&input.AskYesNo{
+			Default: false,
+			Query:   "Move environment folder (SSH key and vault_root_token) to .archive?",
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	if response {
```
JoshVanL: The logic for these responses is not quite right. You need to at least set response to true again before checking auto-approve

pkg/tarmak/cmd.go
```diff
@@ -103,6 +104,118 @@ func (c *CmdTarmak) Destroy() error {
...
+				if err = os.MkdirAll(archivePath, os.ModePerm); err != nil {
+					return err
+				}
+			} else {
+				return fmt.Errorf("failed to get status of path %v", archivePath)
+			}
+		}
+
+		if _, err = os.Stat(environmentArchivePath); err != nil {
+			if os.IsNotExist(err) {
```
JoshVanL: Same here

pkg/tarmak/cmd.go
```diff
@@ -103,6 +104,118 @@ func (c *CmdTarmak) Destroy() error {
...
+				if err = os.Rename(c.Environment().ConfigPath(), environmentArchivePath); err != nil {
+					return err
+				}
+			} else {
+				return fmt.Errorf("failed to get status of path %v", environmentArchivePath)
+			}
+		} else {
```
JoshVanL: Same here

pkg/tarmak/config/config.go
JoshVanL: Can we have a function that lists all clusters in the config?

pkg/tarmak/tarmak.go
```diff
@@ -388,3 +392,17 @@ func (t *Tarmak) verifyImageExists() error {
 
 	return nil
 }
+
+func (t *Tarmak) DestroyClusterMetadata() error {
+	t.log.Infof("removing S3 state and dynamoDB lock for %v", t.Cluster().Name())
```
JoshVanL: This is provider specific, something like `removing tarmak remote state data`

pkg/tarmak/tarmak.go
```diff
@@ -144,24 +144,33 @@ func (t *Tarmak) initializeModules() {
 func (t *Tarmak) initializeConfig() error {
 	var err error
 
+	environmentDestroy := false
+	if t.flags.Environment.Destroy.Name != "" {
```
JoshVanL: We should have a better way to initialize the config. I think we should set the current env to the one to destroy then run this func again

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,169 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+	svc, err := a.S3()
+	if err != nil {
+		return err
+	}
+
+	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
+		Bucket: aws.String(a.RemoteStateName()),
+	})
+
+	if err != nil {
```
JoshVanL: if err == nil {
  return
}

helps with indentation

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,169 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+		ec2err, ok := err.(awserr.Error)
+		if ok && ec2err.Code() == "BucketNotEmpty" {
```
JoshVanL: if !ok || ec2err.Code() != "BucketNotEmpty" {
  return err
}

for indentation again

pkg/tarmak/provider/amazon/tf_remote_state.go
```diff
@@ -385,3 +385,169 @@ func (a *Amazon) initRemoteStateKMS() error {
...
+	}
+
+	result, err := svc.ListObjects(&s3.ListObjectsInput{
+		Bucket: aws.String(a.RemoteStateName()),
+	})
+	if err != nil {
+		return false, err
+	}
+
+	if len(result.Contents) != 0 {
```
JoshVanL: return len(result.Contents) == 0, nil

cmd/tarmak/cmd/environment_destroy.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"errors"
+
+	"github.com/jetstack/tarmak/pkg/tarmak"
+	"github.com/spf13/cobra"
+)
+
+// environmentDestroyCmd handles `tarmak environment destroy`
+var environmentDestroyCmd = &cobra.Command{
+	Use:   "destroy [NAME]",
```
JoshVanL: lowercase name

cmd/tarmak/cmd/environment_destroy.go
```diff
@@ -0,0 +1,35 @@
...
+	Short: "Destroy a environment",
```
JoshVanL: an environment

pkg/tarmak/cmd.go
```diff
@@ -103,6 +104,118 @@ func (c *CmdTarmak) Destroy() error {
...
+			return err
+		}
+	}
+
+	if response {
+		c.log.Info("destroying clusters")
+		for _, cluster := range c.Environment().Clusters() {
+			// We first want to destroy the k8s clusters before the hub
+			if cluster.Name() != c.Environment().Hub().Name() {
+				c.cluster = cluster
```
JoshVanL: We should definitely be checking signals around here

JoshVanL:
@MattiasGees  https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

pkg/tarmak/cmd.go
JoshVanL: Yes sorry, you are quite right

pkg/tarmak/tarmak.go
JoshVanL: I don't really like the fact that we are using Name here in our API which is not being used on the CLI. I think it should be possible to set the current environment, `t.config.CurrentCluster = fmt.Sprintf("%s-", args[0])`
then call `t.initializeConfig()` again within `cmd.DestroyEnvrionment()`

MattiasGees:
@JoshVanL Did a bit of digging into https://github.com/jetstack/tarmak/pull/527#pullrequestreview-172098986

This is because the input sends an error on interruption. https://github.com/jetstack/tarmak/blob/b3c28dfadd0b59bdc8d6329a8eb3ca

If needed I can catch it in the input and return something else, but not sure if it is needed.

MattiasGees:
All fixed.

/unassign
/assign @JoshVanL

pkg/tarmak/config/config.go
```diff
@@ -212,6 +212,13 @@ func (c *Config) Environments() (environments []*tarmakv1alpha1.Environment) {
 	return environments
 }
 
+func (c *Config) AllClusters(environment string) (clusters []*clusterv1alpha1.Cluster) {
```
JoshVanL: Is this used anywhere?

pkg/tarmak/config/config.go
```diff
@@ -281,6 +288,33 @@ func (c *Config) AppendEnvironment(env *tarmakv1alpha1.Environment) error {
...
+	}
+
+	// Set currentCluster to another cluster if it is part of the environment we want to delete
+	currentConfigEnvironment, err := c.CurrentEnvironmentName()
+	if err != nil {
+		return err
+	}
+	// Pick the first found environment/cluster to replace currentCluster
+	if currentConfigEnvironment == environment {
+		c.SetCurrentCluster(fmt.Sprintf("%v-%v", c.conf.Clusters[0].Environment, c.conf.Clusters[0].Name))
```
JoshVanL: We should probably do a range check (>0) on Clusters here

MattiasGees:
/unassign @simonswine 
/assign

Taking this back. Found a bug when destroying a multicluster. In the hub, it doesn't set the socket path for the Tarmak provider.

```
provider "tarmak" {
  socket_path = ""
}
```

MattiasGees:
Fixed it
/unassign
/assign @simonswine

Issue #118, opened 2018-02-20
Test generated terraform code for validity and formatting


simonswine:
We should generate all potential terraform architectures (single-env, multi-cluster env, hub) in CI and make sure that terraform validate/fmt and terraform plan (once we have a single stack) passes

JoshVanL:
@simonswine Guess we can close this now?

simonswine:
Yes this is done in #535

PR #535, opened 2018-09-21
Test generated terraform code

JoshVanL:
Adds fmt and validate on the terraform code when we do the prepare before running any other command.
Improves tarmak CLI output to make clear what steps it's currently on. Useful for the user and debugging to see quickly where something went wrong.
Ensures that tarmak resources are properly cleaned up after running.

related to #118 - this makes adding tests to cli easier

```release-note
Run validate during tarmak preparations
Test terraform fmt during tests
Fixes tarmak resources not being `cleaned up`
```

/assign @simonswine


pkg/tarmak/tarmak.go
```diff
@@ -350,6 +353,9 @@ func (t *Tarmak) Variables() map[string]interface{} {
 }
 
 func (t *Tarmak) Must(err error) {
+	t.Cleanup()
+	plugin.CleanupClients()
+
```
simonswine: This is the wrong method for this. `Must(error)` only handles a fatal error for tarmak. The convention is common in other libraries. Come up with a better method name and call out to the clean up methods and `Must(error)`

pkg/terraform/terraform.go
```diff
@@ -181,10 +189,13 @@ func (t *Terraform) terraformWrapper(cluster interfaces.Cluster, command string,
 			t.socketPath(cluster),
 			stopRpcCh,
 		); err != nil {
-			t.log.Fatalf("error listening to unix socket: %s", err)
+			t.log.Errorf("error listening to unix socket: %s", err)
+			return
```
simonswine: Why are we changing this behaviour. It will create misleading terraform errors

pkg/terraform/terraform.go
```diff
@@ -115,54 +115,62 @@ func (t *Terraform) socketPath(c interfaces.Cluster) string {
...
+			"fmt",
+			"-list=false",
+		},
+		},
+		{log: "validating terraform code", name: "validate", args: []string{
+			"terraform",
+			"validate",
+		},
+		},
+	} {
```
simonswine: What is the benefit of running that every time we plan/apply and destroy.

We should make sure the generated code is always formatted and valid. This should be ensured at PR time through a go test instead (IMHO)

(See my feedback in the review)

simonswine:
/assign @JoshVanL. See my comments the format and validate should be ensured at `go test` time.

A terraform validate would be beneficial to cleaner provider errors, but I don't see any value in the format command

simonswine:
/unassign
/assign @JoshVanL 

Rebase

PR #537, opened 2018-09-24
Bump node-exporter to 0.16.0 and update rules

lostick:
• Bump node-exporter to 0.16.0
• Migrate rules to 0.16.0 format

Most of our infra now exports series with `0.16.0`, we could get rid of `0.15.2` duplication rules altogether if we bump tarmak's node-exporter as well.

**Release note:**
```release-note
Bump node-exporter to 0.16.0 and update rules (breaking changes in metric names https://github.com/prometheus/node_exporter/releases/tag/v0.16.0)
```


simonswine:
Thank you @lostick 

/approve
/lgtm

PR #547, opened 2018-09-26
Fix etcd metrics not being discovered by prometheus

lostick:
• Expose node-exporter metrics for all etcd nodes as an SRV record
• Update prometheus discovery to make sure that etcd node-exporter metrics are correctly scrapped

node-exporter metrics are not being exposed to prometheus, the current config scrapes blackbox exporter metrics instead of the ones coming from node-exporter:
```
- job_name: etcd-nodes-exporter
  scrape_interval: 15s
  scrape_timeout: 10s
  metrics_path: /metrics
  scheme: http
  dns_sd_configs:
  - names:
    - etcd-exporters.xxx.tarmak.local
    refresh_interval: 30s
    type: SRV
    port: 0
```

**Release note**:

```release-note
NONE
```

### DIFF
--- a/puppet/modules/prometheus/README.md
+++ b/puppet/modules/prometheus/README.md
@@ -38,6 +38,11 @@ This module is part of [Tarmak](http://docs.tarmak.io) and should currently be c
 * Type: `Any`
 * Default: `$::prometheus::params::etcd_cluster_exporters`
 
+##### `etcd_cluster_node_exporters`
+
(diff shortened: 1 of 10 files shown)


simonswine:
Thanks @lostick for pointing out that issue. This mistake was introduced once we switched to the DNS discovery. I rather use a simple relabel configuration for that. See #553

PR #563, opened 2018-10-02
Adds `plan --plan-file-store` and `apply plan-file-location`

### LINKED ISSUE #470: Support using the plan output with a subsequent apply

/kind feature

**What you expected to happen**: Terraform supports using the output of a plan in a subsequent apply to ensure you're applying exactly what you've reviewed in a plan - we should support this with Tarmak

### LINKED ISSUE #548: We should allow storing plan output in a file and reusing it on apply

/kind feature

We should allow CI/CD deployments to get the planned step during `cluster plan` and reuse in a later stage in `cluster apply`

JoshVanL:
The plan-file-store flag sets the output location of the plan executable
file.

The apply-file-location sets the file to be used when applying.

Both default to the current default path (current cluster code path).

If a plan-file-location has been set during an apply, the normal plan
that is made before the apply is omitted, only the apply will run.

This enables us to use custom plan executable files in our CI.
We are also able to save plan outputs to arbitrary file locations.

fixes #548 
fixes #470 

```release-note
Adds `tarmak plan --plan-file-store` to sore executable plan output files to arbitrary locations.
Adds `tarmak apply --plan-file-location` to use arbitrary executable plan file for the apply.
- Custom plan files during apply will still execute plan checks.
```

### DIFF
--- a/pkg/tarmak/utils/files.go
+++ b/pkg/tarmak/utils/files.go
@@ -4,6 +4,8 @@ package utils
 import (
 	"os"
 	"path/filepath"
+
+	"github.com/mitchellh/go-homedir"
 )
 
 func EnsureDirectory(path string, mode os.FileMode) error {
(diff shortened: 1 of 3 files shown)


simonswine:
This needs some rebasing similar to #560's `Terraform.dryrun`

/assign @JoshVanL

simonswine:
Sorry again

/unassign
/assign @JoshVanL

PR #565, opened 2018-10-03
Slow down verification of AWS instance types due to rate limit

seanturner83:
Signed-off-by: Sean Turner <sean@secops.ltd>

**When you customise multiple instance types, tarmak fails due to trying to verify these types are available in your region faster than the AWS API will allow**:

```release-note
Slows down verification of AWS instance types so we don't hit the rate limit for these requests
```


simonswine:
Thanks @seanturner83 a solution to that merged via #626

PR #575, opened 2018-10-12
tarmak cluster logs

### LINKED ISSUE #574: Tarmak export logs command
/kind feature

**What happened**:
Often when debugging with someone remotely it is difficult to ensure all logs of interested are collected. This can lead to missing something or increasing the debug loop time to find the problem/solution.

Tarmak should provide a command that goes and fetches the logs which can then be bundled together in a consistent way. This should be done by SSHing into all relevant instances, grabbing journald into a file. Bundle all these files into a tar ball for the user. Exports we should support (not complete):
`tarmak cluster logs vault` - includes consul cluster and helpers
`tarmak cluster logs etcd` - includes backups etc
`tarmak cluster logs worker` - kubelet etc..
`tarmak cluster logs master` - api server etc..

/assign

JoshVanL:
This adds a new cluster sub-command `logs`. The following arguments will be taken as instance pools. This sub-command will go and fetch the logs from every service in each instance of the instance pool for every instance pool given. The logs will then be bundled together into a tar ball for convenient file sharing.

fixes #574 

```release-note
Adds `tarmak cluster logs` for convenient fetching of logs of any instance or target group
```

/assign
TODO: 
- [x] fix cmd hang
- [x] more efficiency


JoshVanL:
/cc @MattiasGees @alljames @simonswine 
Any ideas on how to keep the ssh connection open to improve the speed of fetching logs?
Thoughts on current implementation?

cmd/tarmak/cmd/cluster_logs.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/jetstack/tarmak/pkg/tarmak"
+)
+
+// clusterLogsCmd handles `tarmak clusters logs`
+var clusterLogsCmd = &cobra.Command{
+	Use:   "logs",
```
MattiasGees: Change to `logs [INSTANCE POOL]` to say we need to define an instance pool.

cmd/tarmak/cmd/cluster_logs.go
```diff
@@ -0,0 +1,29 @@
...
+	Short: "Gather logs from an instance pool",
+	Run: func(cmd *cobra.Command, args []string) {
+		t := tarmak.New(globalFlags)
+		defer t.Cleanup()
+
+		if len(args) == 0 {
+			t.Log().Fatal("expecting at least a one instance pool name")
+		}
+
+		t.CancellationContext().WaitOrCancel(t.NewCmdTerraform(args).Logs)
```
MattiasGees: Why `NewCmdTerraform`? I don't feel this belongs here as the logs have nothing to do with Terraform.

JoshVanL:
- Sane defaults
  - Retrieve logs for 24h
  - Control plan + Etcds + Bastion + Vault default targets
- Look into the rotation time of journald
- Allow specifc instance or instances (similar name like in cluster ssh instance)
- Use a single journalctl -o json stream per instance
- Filter then per service / kernel / target on the tarmak host

cmd/tarmak/cmd/cluster_logs.go
JoshVanL: I agree however neither is SSH(). I like using this struct for consistency and using the context etc. Perhaps we should rename it to CmdTarmak however this doesn't make much sense being in terraform.go
Thoughts?

JoshVanL:
/unassign
/assign @simonswine @MattiasGees 

Logs are now json streamed and piped straight to file to be bundled together.
All hosts in a group are gathered concurrently.
Sane defaults for since + until.

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+	ssh    interfaces.SSH
+	ctx    interfaces.CancellationContext
+	log    *logrus.Entry
+
+	path    string
+	since   string
+	until   string
+	targets []string
+
+	mu       sync.Mutex
```
simonswine: Make clearer what that lock is protecting

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+	wg       sync.WaitGroup
+	hosts    []interfaces.Host
+	tmpDir   string
+	tmpFiles map[string]*os.File
```
simonswine: I think you should make sure that this has it's own lock protecting read/writes to the map

pkg/tarmak/logs/logs.go
simonswine: in its name also a comment

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+	TargetGroups = []string{
+		"bastion",
+		"vault",
+		"etcd",
+		"worker",
+		"control-plane",
+	}
+)
+
+type Logs struct {
```
simonswine: Add some comments to make clear what var are for

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+
+	for readerScanner.Scan() {
+		r := readerScanner.Bytes()
+
+		if r == nil || len(r) == 0 {
+			continue
+		}
+
+		entry := new(SystemdEntry)
+		if err := json.Unmarshal(r, entry); err != nil {
```
simonswine: use json.NewDecoder and a decode loop instead of a scanner + individual decodes

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+			l.mu.Lock()
+			err = fmt.Errorf("failed to unmarshal entry [%s]: %s", r, err)
+			result = multierror.Append(result, err)
+			l.mu.Unlock()
```
simonswine: Keep the are with the lock separate from other code (new lines before and after). also keep it minimal (i.e.) without the Errorf

cmd/tarmak/cmd/cluster.go
```diff
@@ -49,6 +57,32 @@ func clusterDestroyFlags(fs *flag.FlagSet) {
...
+func clusterLogsFlags(fs *flag.FlagSet) {
+	store := &globalFlags.Cluster.Logs
+	now := time.Now()
+
+	fs.StringVar(
+		&store.Path,
+		"path",
+		utils.DefaultLogsPathPlaceholder,
+		"target tar ball path",
+	)
```
simonswine: Maybe we default to the current working directory, config files for tarmak should not be polluted

cmd/tarmak/cmd/cluster_logs.go
```diff
@@ -0,0 +1,42 @@
...
+// clusterLogsCmd handles `tarmak clusters logs`
+var clusterLogsCmd = &cobra.Command{
+	Use: "logs [target groups]",
+	Long: fmt.Sprintf(
+		"Gather logs from a list of instances or target groups %s",
+		logs.TargetGroups,
+	),
+	Short: "Gather logs from a list of instances or target groups",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
```
simonswine: Some better validation of targets would be great.

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	l.ssh = l.tarmak.SSH()
+	l.since = fmt.Sprintf(`"%s"`, flags.Since)
+	l.until = fmt.Sprintf(`"%s"`, flags.Until)
+	l.tmpFiles = make(map[string]*os.File)
+
+	switch group {
+	case "control-plane":
+		l.targets = []string{"master", "etcd"}
```
simonswine: Maybe we define vault as being part of the control-plane?

cmd/tarmak/cmd/cluster.go
MattiasGees: yeah I also think current working directory is a better solution.

alljames: :+1:

pkg/tarmak/interfaces/interfaces.go
```diff
@@ -208,6 +208,11 @@ type SSH interface {
 	PassThrough([]string)
 	Tunnel(hostname string, destination string, destinationPort int) Tunnel
 	Execute(host string, cmd string, args []string) (returnCode int, err error)
+	ExecuteWithWriter(host string, cmd string, args []string, stdout io.Writer) (returnCode int, err error)
+}
+
+type Logs interface {
+	Gather(group string, flags tarmakv1alpha1.ClusterLogsFlags) error
```
alljames: or `Aggregate` ?

cmd/tarmak/cmd/cluster_logs.go
JoshVanL: I've done more validation further down. I don't want to call ListHosts() here again because it's so expensive

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,371 @@
...
+)
+
+const (
+	logTimeLayout = "2006-01-02 15:04:05" // journalctl time format
+	timeLayout    = "Jan _2 15:04:05"     // journald style layout
+)
+
+var (
+	// target groups of one or more instance groups
+	TargetGroups = []string{
```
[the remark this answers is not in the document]
JoshVanL: Can put it in constants if you like. Going to also namespaces that more -> utils/constants/constants.go

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,371 @@
...
+
+	if flags.Until == utils.DefaultLogsUntilPlaceholder {
+		l.until = fmt.Sprintf("'%s'", now.Format(logTimeLayout))
+	} else {
+		l.until = fmt.Sprintf("'%s'", flags.Until)
+	}
+
+	switch group {
+	case "control-plane":
+		l.targets = []string{"vault", "etcd", "master"}
```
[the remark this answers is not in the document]
JoshVanL: :+1:

pkg/tarmak/logs/logs.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+
+		l.wg.Add(1)
+		go hostGatherFunc(a)
+	}
+
+	l.wg.Wait()
+
+	if result != nil {
+		return result.ErrorOrNil()
```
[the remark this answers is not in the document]
JoshVanL: if result != nil then it can't be nil :stuck_out_tongue:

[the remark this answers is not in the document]
MattiasGees: Maybe remove the `.ErrorOrNil()` as it can never be Nil to not cause any confusion.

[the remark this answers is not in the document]
JoshVanL: result is type  *multierror.Error so .ErrorOrNil() returns type error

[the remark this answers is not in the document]
MattiasGees: you can just do `return result` as `multiError.Error` implements `error` I think.

MattiasGees:
Looks good.

/approve
/lgtm
/unassign
/assign @simonswine

simonswine:
Not too sure how much work it is, but would be great to get into 0.6

/assign @JoshVanL 
/unassign

simonswine:
I think almost there, i find this behaviour odd:

```
./_output/tarmak cluster logs --path test.tar.gz master vault                                                                                                                                                                                     1 ↵
INFO[0000] fetching logs from target 'master'            app=tarmak
INFO[0001] fetching journald logs from instance 'master-3'  app=tarmak
INFO[0001] fetching journald logs from instance 'master-1'  app=tarmak
INFO[0001] fetching journald logs from instance 'master-2'  app=tarmak
INFO[0001] new connection to bastion host successful     app=tarmak
INFO[0004] logs bundle written to '/home/simon/.golang/packages/src/github.com/jetstack/tarmak/test.tar.gz'  app=tarmak
INFO[0004] fetching logs from target 'vault'             app=tarmak
INFO[0004] fetching journald logs from instance 'vault-3'  app=tarmak
INFO[0004] fetching journald logs from instance 'vault-2'  app=tarmak
INFO[0004] fetching journald logs from instance 'vault-1'  app=tarmak
INFO[0005] logs bundle written to '/home/simon/.golang/packages/src/github.com/jetstack/tarmak/test.tar.gz'  app=tarmak
INFO[0005] tarmak performed all tasks successfully
```

They should not overwrite each other, i guess

`control-plane` and `master etcd vault` should do the exact same

simonswine:
Thanks @JoshVanL, looking good now

/approve
/lgtm

PR #578, opened 2018-10-15
Rate limit aws verify (to avoid RequestLimitExceeded error)

alljames:
When a large number of AWS 'describe' API calls are made within a short period of time, AWS can return a RequestLimitExceeded error. This PR deals with this by implementing a retry with backoff.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #565 

Uses backoff library instead of a hard-coded sleep
github.com/cenkalti/backoff - this library is already used elsewhere in the Tarmak source code 

**Release note**:
```release-note
Retry backoff implemented when verifying instance types to prevent RequestLimitExceeded error from AWS API
```

### DIFF
--- a/pkg/tarmak/provider/amazon/amazon.go
+++ b/pkg/tarmak/provider/amazon/amazon.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+        "time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -585,6 +586,8 @@ func (a *Amazon) verifyInstanceTypes() error {
 		if err := a.verifyInstanceType(instanceType, instance.Zones(), svc); err != nil {
 			result = multierror.Append(result, err)
 		}
+
+		time.Sleep(200 * time.Millisecond)
 
 	}
 
--- a/pkg/tarmak/provider/amazon/amazon.go
+++ b/pkg/tarmak/provider/amazon/amazon.go
@@ -7,7 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-        "time"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -576,16 +582,31 @@ func (a *Amazon) verifyInstanceTypes() error {
...
-			result = multierror.Append(result, err)
+		verifyInstanceTypeFunc := func() error {
+			err = a.verifyInstanceType(instanceType, instance.Zones(), svc)
+			if err != nil {
+				// return an error if we're exceeding AWS API request limits
+				if strings.Contains(err.Error(), "RequestLimitExceeded") {
+					return err
+				}
+			}
+			return nil
```
JoshVanL: This is not quite the functionality that we want.

Backoff will continue to try the function so long as the function returns an error and the context - in this case a max try - is good. In it's current state, an error that is not a rate limit will return fine when this is not the case. Instead here we should use a context where an error that is not a rate limit should return with that error immediately.

```
ctx, cancelRetries := context.WithCancel(context.Background()) 
b := backoff.WithContext(expBackoff, ctx)
```

pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -576,16 +582,31 @@ func (a *Amazon) verifyInstanceTypes() error {
...
 			return err
 		}
 
-		if err := a.verifyInstanceType(instanceType, instance.Zones(), svc); err != nil {
-			result = multierror.Append(result, err)
+		verifyInstanceTypeFunc := func() error {
+			err = a.verifyInstanceType(instanceType, instance.Zones(), svc)
+			if err != nil {
+				// return an error if we're exceeding AWS API request limits
+				if strings.Contains(err.Error(), "RequestLimitExceeded") {
```
JoshVanL: This error string would be nice as a constant to be used elsewhere

simonswine:
Solution to that merged via #626

PR #579, opened 2018-10-15
Enables additional policies for bastion and vault instance pools

### LINKED ISSUE #519: Support additional IAM policies for non-templated instance pools

/kind bug

**What happened**: The [docs](https://docs.tarmak.io/user-guide.html#instance-pool) state that additional IAM policies can be added to all instance pools, but in fact this only works for [templated instance pools](https://github.com/jetstack/tarmak/blob/master/terraform/amazon/templates/instance_pools/instance_pools.tf.template#L25) and [Jenkins](https://github.com/jetstack/tarmak/blob/master/terraform/amazon/templates/instance_pools/instance_pools.tf.template#L51).

**What you expected to happen**: We should support this for Bastion and Vault instance pools too. This will require creating a role in Terraform for the bastion.

**How to reproduce it (as minimally and precisely as possible)**: Try adding an additional IAM policy to a Vault instance pool

JoshVanL:
Additional IAM policies from the tarmak config were not enabled for vault and bastion instance pools. This PR creates a instance profile to the bastion instance to attach policies to. 
Bastion and vault instance pools now attach policies from arns in the config

fixes #519 

```release-note
Bastion and Vault instance pools now support additional policies declared in the config
```

/assign @simonswine @dippynark

### DIFF
--- a/pkg/apis/cluster/v1alpha1/cluster.go
+++ b/pkg/apis/cluster/v1alpha1/cluster.go
@@ -72,8 +72,8 @@ type Cluster struct {
 
 // ClusterAmazon offers Amazon-specific settings for that instance pool
 type ClusterAmazon struct {
-	// This fields contains ARNs for additional IAM policies to be added to this
-	// instance pool
+	// This fields contains ARNs for additional IAM policies to be added to
+	// this instance pool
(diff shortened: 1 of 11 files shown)


simonswine:
Needs rebasing

/assign @JoshVanL 
/unassign @simonswine @dippynark

PR #587, opened 2018-10-19
Move tarmak sock to /tmp and ensure cleanup

### LINKED ISSUE #586: Move tarmak socket to its own path

>
> /kind bug
/kind feature

**What happened**:
Sockets are limited in path length. The socket now gets created inside the config directory of Tarmak. It is possible that that config directory path will be too long. 

I propose to create the tarmak.sock in another path than the Tarmak config directory. This will not limit our users where to store the Tarmak config directory.
A possible location we could use is `/tmp`. 
When we move it to a tmp folder, we should think about creating a unique socket name. This will be needed when we run Tarmak at the same time for different clusters. This can happen in CI/CD environments.

We did something similar for the controlpath in the SSH multiplexing. 
https://github.com/jetstack/tarmak/pull/179

JoshVanL:
There is a maximum limit to the path name of sockets. This caused problems with config files in deep file paths. This moves the socket to /tmp so is never too long. This also ensures that all resources created in /tmp are now deleted at the end of run.

fixes #586 

```release-note
Move tarmak terraform provider socket to /tmp
```

/assign @MattiasGees 
/cc @MattiasGees @simonswine

### DIFF
--- a/cmd/tarmak/cmd/cluster_force-unlock.go
+++ b/cmd/tarmak/cmd/cluster_force-unlock.go
@@ -12,11 +12,8 @@ var clusterForceUnlockCmd = &cobra.Command{
 	Short: "Remove remote lock using lock ID",
 	Run: func(cmd *cobra.Command, args []string) {
 		t := tarmak.New(globalFlags)
-		defer t.Cleanup()
 
-		forceUnlockCmd := t.NewCmdTerraform(args)
-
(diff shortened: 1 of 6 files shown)


MattiasGees:
Tested and works perfectly.

/lgtm
/approve
/unassign
/assign @simonswine

simonswine:
/approve
/lgtm

Thanks Josh that sounds great

PR #589, opened 2018-10-19
Auto-generate cobra CLI documentation for the site

### LINKED ISSUE #588: Generate cobra docs for all commands
/kind feature

**What happened**:
Instead of manually updating the CLI documentation we should generate it and incorporate it in CI, like we do with the API documentation.
https://github.com/spf13/cobra/blob/master/doc/md_docs.md

JoshVanL:
Automates generating CLI documentation in the site and ensures it's up to date in CI.

fixes #588 

```release-note
Auto-generate CLI documentation
```


docs/generated/cmd/tarmak/tarmak_clusters_set-current.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Synopsis
+~~~~~~~~
+
+
+Set current cluster in config
+
+::
+
+  tarmak clusters set-current [flags]
```
MattiasGees: I don't know if it is possible with the generated docs, but it would be nice to have also some generated docs for the argument. This is the case now https://docs.tarmak.io/design_cli_ux.html#set-current

MattiasGees: I don't know if it is possible with the generated docs, but it would be nice to have also some generated docs for the argument. This is the case now https://docs.tarmak.io/design_cli_ux.html#set-current

simonswine:
Needs rebase 

/assign @JoshVanL 
/unassign

Makefile
```diff
@@ -189,11 +189,16 @@ verify_boilerplate:
...
+	dep ensure -no-vendor -dry-run -v
+
+verify_gen_docs: verify_reference_docs verify_cmd_docs
+
 verify_reference_docs: $(BINDIR)/node
 	$(HACK_DIR)/verify-reference-docs.sh
 
-verify_vendor: $(BINDIR)/dep
-	dep ensure -no-vendor -dry-run -v
+verify_cmd_docs:
```
MattiasGees: Shouldn't we also add a `go_cmd_docs_gen` command?

JoshVanL: Not sure what you mean

JoshVanL: I see what you mean now, yes that makes sense

PR #590, opened 2018-10-22
Update cobra shorts with arguments for docs gen

JoshVanL:
We are going to automate generating command documentation. As such, we should add some more information about what arguments command take.

```release-note
NONE
```

/assign @MattiasGees @simonswine

### DIFF
--- a/cmd/tarmak/cmd/cluster_set-current.go
+++ b/cmd/tarmak/cmd/cluster_set-current.go
@@ -8,14 +8,14 @@ import (
 )
 
 var clusterSetCurrentCmd = &cobra.Command{
-	Use:   "set-current",
+	Use:   "set-current [environment-cluster]",
 	Short: "Set current cluster in config",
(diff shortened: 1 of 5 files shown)


cmd/tarmak/cmd/cluster_set-current.go
```diff
@@ -8,14 +8,14 @@ import (
...
 var clusterSetCurrentCmd = &cobra.Command{
-	Use:   "set-current",
+	Use:   "set-current [environment-cluster]",
 	Short: "Set current cluster in config",
 	Run: func(cmd *cobra.Command, args []string) {
 		t := tarmak.New(globalFlags)
 		defer t.Cleanup()
 
 		if len(args) != 1 {
-			t.Log().Fatalf("Only one cluster can be set as current cluster")
```
simonswine: Fatalf without arguments

PR #596, opened 2018-10-24
Fix broken cluster API link

JoshVanL:
```release-note
NONE
```

/assign @simonswine @MattiasGees


simonswine:
@JoshVanL thanks for finding and fixing!

Can we make sure that we only assign issue to exactly one person?

/unassign @MattiasGees 
/lgtm
/approve

PR #602, opened 2018-10-26
Remove myself from OWNERS

charlieegan3:
I don't want to be auto-assigned PRs any more

```release-note
NONE
```


munnerz:
You'll need one of @simonswine @JoshVanL or @dippynark to add /approve (although I'll do it below too, just to double check that it doesn't work)

/lgtm
/approve

PR #604, opened 2018-10-26
Better image build control

### LINKED ISSUE #599: Better control over `cluster image build`
/kind feature

Right now it seems that `image build` will just re-build all images that are listed in the current cluster instance pools. We should provide a way to more finely control this behavior.

IMHO, default behavior should not re-build images that are already existing.
This can be overwritten with an override/force flag.
Enable to select a subset of images to build.

JoshVanL:
As we are adding more base images for different instance roles we should have better control over building these images. This adds the ability to only build missing images, specific images or all images. This pairs with the images destroy sub-command.

fixes #599 

```release-note
Add better `cluster images build` CLI behavior
```

/assign @MattiasGees


MattiasGees:
works

/lgtm
/approve
/unassign
/assign @simonswine

JoshVanL:
@simonswine good catch, forgot to check for matching encryption

JoshVanL:
@simonswine good catch, forgot to check for matching encryption

PR #605, opened 2018-10-30
Adds restart on failure to consul service

JoshVanL:
If for whatever reason consul exited, the systemd unit would not attempt to restart it. 
This PR adds a 10 second cool off before restarting, in-line with our upstream.

```release-note
Adds restart on failure to the consul service
```

### DIFF
--- a/terraform/amazon/modules/vault/templates/vault_user_data.yaml
+++ b/terraform/amazon/modules/vault/templates/vault_user_data.yaml
@@ -115,6 +115,8 @@ write_files:
        -server \
        -config-dir=/etc/consul \
        -advertise=${private_ip}
+    Restart=on-failure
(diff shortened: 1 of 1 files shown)


simonswine:
@JoshVanL Can we also add that to the vault-unsealer service?

/unassign
/assign @JoshVanL

JoshVanL:
@simonswine It already has restart=always
https://github.com/jetstack/tarmak/blob/94d1b73a8b1115445e411fe8b60a6736528e9083/terraform/amazon/modules/vault/templates/vault_user_data.yaml#L149

simonswine:
You are right, but it's not in the last lines as in the other unit :joy: 

/approve
/lgtm

PR #606, opened 2018-10-30
Selects a ready vault instances instead of vault-1

JoshVanL:
Tarmak will select the first vault instance to create a tunnel to the cluster. This PR selects any of the tunnels when one is ready.

```release-note
Select any vault tunnel once one is ready
```

### DIFF
--- a/pkg/tarmak/vault/vault.go
+++ b/pkg/tarmak/vault/vault.go
@@ -3,6 +3,7 @@ package vault
 
 import (
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


pkg/tarmak/vault/vault.go
```diff
@@ -214,7 +217,24 @@ func (v *Vault) VerifyInitFromFQDNs(instances []string, vaultCA, vaultKMSKeyID,
 		return err
 	}
 
-	cl := tunnels[0].VaultClient()
+	var cl *vault.Client
+	readyTunnelFunc := func() error {
+		for _, t := range tunnels {
+			if t.Status() != VaultStateErr {
```
simonswine: I know this is just a backport, but I think we might want to check for more cases here. (Sealed, not initialised)

JoshVanL: This is done afterwards. This means that we can keep using the same tunnel for these checks instead of creating them over again, when the cluster is already in a responsive state (initialized, sealed etc.)

PR #625, opened 2018-11-13
Adds Jenkins instance pool tests to tarmak_test.go

JoshVanL:
Includes Jenkins instance pool into tarmak_test.go

```release-note
NONE
```

### DIFF
--- a/pkg/tarmak/config/cluster.go
+++ b/pkg/tarmak/config/cluster.go
@@ -43,6 +43,15 @@ func NewClusterMulti(environment string, name string) *clusterv1alpha1.Cluster {
 	return c
 }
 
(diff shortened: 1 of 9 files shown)


simonswine:
My concern is/was we did not call `terraform validate` on those stacks so the error was not catched.

You can have a look at my code but that will fail as `terraform init` blocks as it wants some connection to the state bucket

https://github.com/jetstack/tarmak/commit/1f00bd849459a0d910f3e4c63fdc1effc5d98a41

JoshVanL:
@simonswine Guess we should have some flag for controlling off/on line tests to skip or run this

simonswine:
I guess maybe if we don't configure the remote state (and use local) it should work, I will create an inssue for that 

/assign @simonswine

PR #626, opened 2018-11-13
Adds backoff to describe reserved instance offerings

JoshVanL:
Adds a back-off to AWS Describe Reserved Instance Offerings call in cases of rate limiting. If the error is anything else then we exit right away.

```release-note
NONE
```

/assign @simonswine

### DIFF
--- a/pkg/tarmak/provider/amazon/amazon.go
+++ b/pkg/tarmak/provider/amazon/amazon.go
@@ -2,11 +2,14 @@
 package amazon
 
 import (
+	"context"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


pkg/tarmak/provider/amazon/amazon.go
```diff
@@ -604,7 +608,34 @@ func (a *Amazon) verifyInstanceType(instanceType string, zones []string, svc EC2
...
+	var response *ec2.DescribeReservedInstancesOfferingsOutput
+	describeInstanceOfferingFunc := func() error {
+		var err error
+		response, err = svc.DescribeReservedInstancesOfferings(request)
+		if err == nil {
+			return nil
+		}
+
+		// if err and it's a rate limiting error, try again
+		if strings.Contains(err.Error(), consts.AmazonRateLimitErr) {
```
simonswine: Can we handle the aws error like that: https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html

`awserr.Code()` should contain your AmazonRateLimitErr

PR #632, opened 2018-11-15
Adds `tarmak cluster kubeconfig` subcommand

### LINKED ISSUE #279: Provide a kubeconfig subcommand for clusters

JoshVanL:
Adds `tarmak cluster kubeconfig` sub command to verify and create the kubeconfig file.
Supports use of eval "$(tarmak cluster kubeconfig)" to set the KUBECONFIG environment variable.
Can set output path with --path flag. Default is default kubeconfig path.
Can set kubeconfig to point to the public API endpoint.

fixes #279 

```release-note
Adds `tarmak cluster kubeconfig`
```

/assign @simonswine


cmd/tarmak/cmd/cluster.go
```diff
@@ -95,6 +95,25 @@ func clusterFlagDryRun(fs *flag.FlagSet, store *bool) {
 	)
 }
 
+func clusterKubeconfigFlags(fs *flag.FlagSet) {
+	store := &globalFlags.Cluster.Kubeconfig
+
+	fs.BoolVar(
+		&store.PublicAPIEndpoint,
+		"public-api-endpoint",
+		false,
```
simonswine: This should be true for clusters where there is a public API endpoint available, false if not

Both values should be able to be overriden

JoshVanL:
/unassign
@simonswine behavior should be better now

simonswine:
Thank you a few more thoughts:

- Should we have the same public-api-server flag behaviour on kubectl (would probably mean we use it as a global flag instead)?
- I would like to have that example usage in the docs:

```
eval "$(tarmak cluster kubeconfig)"
kubectl get nodes
helm install
```

/assign @JoshVanL 
/unassign

JoshVanL:
@simonswine Yep, guess that will have to be global :+1:

simonswine:
I think this is almost ready:

```
tarmak kubeconfig --public-api-endpoint=true
DEBU[0000] check SSH connection to bastion cmd=[ssh -F /home/christian/.tarmak-dev/csire-cluster/ssh_config bastion /bin/true]  app=tarmak destination=api.csire-cluster.tarmak.local
DEBU[0000] start tunnel cmd=[ssh -F /home/christian/.tarmak-dev/csire-cluster/ssh_config -O forward -L127.0.0.1:41351:api.csire-cluster.tarmak.local:6443 bastion]  app=tarmak destination=api.csire-cluster.tarmak.local
DEBU[0000] trying to connect to https://api.csire-cluster.develop.tarmak.org  app=tarmak
DEBU[0000] connected to Kubernetes API v1.11.3           app=tarmak
KUBECONFIG=/home/christian/.tarmak-dev/csire-cluster/kubeconfig
```

I am not too sure why there is this tunnel start line. Ideally we would not need any tunnel for public APIs (apart from getting the vault cert). (It is the same for kubectl)

Are we starting a tunnel or not?

/assign @JoshVanL 
/unassign

simonswine:
That is it thank you!

PR #639, opened 2018-11-22
Kube bench proposed security fixes

simonswine:
```release-note
Disable profiling endpoints for control plane components
Enable service account lookup in etcd
Disable kubelets read-only ports
Hardening TLS configuration for Kubernetes >= 1.10
Do not repair malformed updates in apiserver
Fix insecure API server port enabled in >= 1.11
```


simonswine:
/assign @JoshVanL 

Can you give that a review please?

PR #643, opened 2018-11-30
Adds management of ssh_known_hosts file proposal

JoshVanL:
As we are replacing the ssh command line tool for all programmatic use cases in favor of an in package solution, we have discovered some problems managing the ssh_known_hosts file correctly. This proposal aims to address these. 

```release-note
Management of SSH known_hosts file Proposal
```

### DIFF
--- a/docs/proposals/ssh-known_hosts.rst
+++ b/docs/proposals/ssh-known_hosts.rst
@@ -101,6 +101,9 @@ Notable items
 
 Public DSA keys will not be tagged.
 
+Care must be taken to ensure that Terraform does not override tags set by the
+Lambda function.
+
 A start has been made on the code for the Lambda function:
 
 .. code-block:: go
@@ -206,6 +209,17 @@ A start has been made on the code for the Lambda function:
   func main() {
   	lambda.Start(HandleRequest)
   }
+
+
+Limitations
+-----------
+Whilst we can restrict permissions of access for both the Lambda function and
+EC2 instances, we do not have a cryptographic signature of the public keys
+coming from the EC2 instance.
+
+Whilst using Hashicorp's Vault to set up an SSH CA for the environment would be
+advantageous, bootstrapping this process requires SSH connections from the client
+to EC2 instances. This rules out this as an option.
 
 Out of scope
 ------------
--- a/docs/proposals/ssh-known_hosts.rst
+++ b/docs/proposals/ssh-known_hosts.rst
@@ -1,0 +1,93 @@
+.. vim:set ft=rst spell:
+
+Managing SSH Known Hosts
+========================
+
+This proposal suggests a method to more securely manage and ensure
+authentication of remote hosts through the SSH protocol within Tarmak
+environments on AWS.
+
+Background
+----------
+
+Currently we solely use the external OpenSHH SSH command line tool to connect to
+remote instances on EC2 for both interactive shell sessions as well as
+tunnelling and proxy commands for other services, including, connecting to the
+Vault cluster and the private Kubernetes API server endpoint. Currently in
+development is the replacement of our programmatic use cases of SSH in favour of
+the in package Go solution, a choice stemming from pain points developing more
+sophisticated utility functions for Tarmak and the desire for improvements in
+control of connections to remote hosts.
+
+During development of this replacement it became clear that proper care must be
+taken during authentication of host public keys during connection and manual
+management of our ``ssh_known_hosts`` cluster file. Our current implementation
+allows OpenSSH to maintain this file however, does not exit with an error if
+public keys do not match due to the flag ``StrictHostKeyChecking`` set to
+``no``. Not only does a miss-match in public keys not cause an error, the
+population of known public keys on different authenticated machines to the same
+EC2 hosts will always use the hosts presented public key, meaning the set of
+public keys could potentially be different for users accessing the same cluster.
+
+Objective
+---------
+
+By implementing stricter enforcement of the ``ssh_known_hosts`` file and passing
+it's management to Tarmak, we can improve the security of SSH connections to
+remote hosts. The key high level points to achieving this is as follows:
+
+ - Disable writes from the OpenSSH command to the ``ssh_known_hosts`` file and
+   enforce strict checking.
+ - Enforce that our in package implementation of SSH connections adheres to this
+   file also.
+ - Collect public keys during instance start up that are then stored, tightly
+   coupled with that host. These keys are able to be used as a source of truth
+   for other authenticated users attempting to connect to remote hosts on the
+   cluster that have empty or an incomplete ``ssh_known_hosts`` file.
+
+Changes
+-------
+
+Firstly, we must restrict the OpenSSH command line tool from editing the
+``ssh_known_hosts`` file and strictly enforce it by updating the generator for the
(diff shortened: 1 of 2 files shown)


docs/proposals/ssh-known_hosts.rst
```diff
@@ -0,0 +1,93 @@
...
+
+In order to create a source of truth for each host's public key, each instance
+will have it's public key attached as a tag, shortly after boot time like the
+following:
+
++------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| public_key | ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBPF+xkIGMUNVI0gElRaTLjfA4QMN/XGJhHswDyv59DNSOtG3KwZvDF3YkAb0PkTQAYo8N5fxoKqimGugOAaefPc= |
++------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+
+The population of this tag will occur during a Terraform apply in which the
```
munnerz: Doesn't tarmak also use 'autoscaling groups' to autoprovision instances? How will these have the tag populated?

docs/proposals/ssh-known_hosts.rst
```diff
@@ -0,0 +1,93 @@
...
+public keys could potentially be different for users accessing the same cluster.
+
+Objective
+---------
+
+By implementing stricter enforcement of the ``ssh_known_hosts`` file and passing
+it's management to Tarmak, we can improve the security of SSH connections to
+remote hosts. The key high level points to achieving this is as follows:
+
+ - Disable writes from the OpenSSH command to the ``ssh_known_hosts`` file and
```
munnerz: Would it be easier for tarmak to maintain an 'in memory' ssh_known_hosts map (that could be 'generated' when tarmak needs to SSH to instances), and then add a subcommand `tarmak get known-hosts` or something in case users want to access/get a copy of the file?

Perhaps there's some other reason why you'd rather share a single file for this that I've missed! 😄

JoshVanL: I quite like having this file around as it makes it easier for using ssh without tarmak - we also keep the SSH config around. Same idea as having the kubeconfig in file to use kubectl without tarmak.

munnerz: Sure thing - I just thought it might be simpler as you won't have to deal with user-introduced unknowns!

docs/proposals/ssh-known_hosts.rst
simonswine: This is the tricky bit here. In theory the nodes could self update it's tags. But I think the AWS IAM Policies can't be configured that tight to allow only self update. This also has some issues, once the node is overtaken and other tags (e.g. tarmak_role) are no longer safe.

Another problem is that the amount of tags to be stored on an AWS instance cannot exceed 255 chars. Also we have more than just the ecdsa host key.

MattiasGees: I also think we have to look further than AWS. Not every provider will have support for tags. I think we should come up with a universal solution for this.

JoshVanL: While I agree, IMO taging instances on AWS would be the best way to couple the public keys with the instance life cycle. If we keep the SSH config with,
`HostKeyCallback: ssh.tarmak.Provider().HostKeyCallBack,`
then each provider can have it's specific implementation. 
This is assuming all targeted providers will have some kind of instance metasdata we can pull and push from

MattiasGees:
TBH I don't like the solution with the lambda function. This is really AWS specific and I know we can always have a specific implementation. But IMHO this is reinventing the wheel. 
Why don't we use Hashicorp Vault for generating a CA that can be used for the host verification key? Both the user and server will need to trust that CA and can use that CA to verify the connection.

With this we would end up with a solution that works everywhere. 

Some more information: https://jameshfisher.com/2018/03/16/how-to-create-an-ssh-certificate-authority.html

docs/proposals/ssh-known_hosts.rst
```diff
@@ -0,0 +1,211 @@
...
+``ssh_known_hosts`` file and strictly enforce it by updating the generator for the
+``ssh_config`` file. This enables Tarmak to take control of the ``ssh_known_hosts``
+file management.
+
+In order to create a source of truth for each host's public key, each instance
+will have it's public key's attached as a tags, shortly after boot time like the
+following:
+
++-------------------------+---------------------------------------------------------------------------+
+| PublicKey_ssh-ed25519_0 | AAAAC3NzaC1lZDI1NTE5AAAAIE90XYYm6GSDlNGejM+aY5dZEe5vK4XyU++89WdGJcDc==EOF |
```
simonswine: don't really like the format here: maybe we do something like that:
`tarmak.io/ssh-host-ed255519-pub-0`
`tarmak.io/ssh-host-ecdsa-pub-0`
`tarmak.io/ssh-host-rsa-pub-0`

I suggest also we don't worry about dsa

docs/proposals/ssh-known_hosts.rst
```diff
@@ -0,0 +1,211 @@
...
+Changes
+-------
+
+Firstly, we must restrict the OpenSSH command line tool from editing the
+``ssh_known_hosts`` file and strictly enforce it by updating the generator for the
+``ssh_config`` file. This enables Tarmak to take control of the ``ssh_known_hosts``
+file management.
+
+In order to create a source of truth for each host's public key, each instance
+will have it's public key's attached as a tags, shortly after boot time like the
```
simonswine: Also some considerations regarding if we can make sure terraform is not removing those tags every time we run it.

This is esp important for aws_instance resources. Maybe https://www.terraform.io/docs/configuration/resources.html#ignore_changes is good enough

docs/proposals/ssh-known_hosts.rst
```diff
@@ -0,0 +1,211 @@
...
++-------------------------+---------------------------------------------------------------------------+
+| PublicKey_ssh-ed25519_0 | AAAAC3NzaC1lZDI1NTE5AAAAIE90XYYm6GSDlNGejM+aY5dZEe5vK4XyU++89WdGJcDc==EOF |
++-------------------------+---------------------------------------------------------------------------+
+
+The population of these tags will happen at boot time for all instances,
+regardless of whether they have been created from a direct Terraform apply or
+via an Amazon Auto Scaling Group. At execution time, Wing - present on every
+instance - will invoke an Amazon Lambda function for Instance Tagging. Passed to
+this function will be a collection of the instances public keys, it's Amazon
+identity document and matching PKCS7 document.
```
simonswine: While the instance document provides some assurances that it is an actual instance from AWS we should not done the limitations of that in the proposal:

- There is no cryptographic signature that the public keys come actually from the host that is sending the request
- The lambda function has all the information acting like the instance to someone else with that document (e.g. logging into vault with the same method).  

The instance document could be coming form another instance and

simonswine:
Thank you josh. I am having some issues with it's portability as well (concerns Mattias mentioned), but I can't really see any short term work arounds. Esp at bootstrap it's really hard to do that correct

/approve
/lgtm

PR #658, opened 2018-12-11
Fix latest kernel packer image

simonswine:
```release-note
Make sure centos-puppet-agent-latest-kernel is booting into the right kernel
```


simonswine:
@JoshVanL this seems to be something I introduced through #649

alljames:
Have tested this by specifying `centos-puppet-agent-latest-kernel` in `tarmak.yaml`, building an AMI and spinning up a Tarmak cluster.  I can confirm that SSHing into each instance and running `uname -a` shows that all instances are running the `4.19.8-1.el7.elrepo.x86_64` kernel, i.e. the [latest stable kernel](https://www.kernel.org/)

/lgtm

PR #663, opened 2018-12-17
Use machine API types

JoshVanL:
This PR implements the Machine API types taken from the Kubernetes Cluster API project.
This introduces MachineDeployments and MachineSets as well as Machine, essentially renamed from Instances.
Two new controllers are added to the API server that manage the behavior of Machine deployments and sets.

Every instance pool now has a corresponding deployment, and consequently a set, created to manage the machines in that pool.

This PR helps to implement ground work to further extend wing features such as puppet dry runs.

**NOTE**: wing-dev-mode _must_ be run with this branch since wing and it's types have been heavily modified.

```release-note
Change wing API types to use ones similar to Kubernetes Cluster API
```


JoshVanL:
Should be ready for a first round of reviews
/assign @simonswine

simonswine:
/hold

I would not like to include this 0.6 release

PR #494, opened 2018-09-06
Use wing + puppet to bootstrap vault/consul instances

JoshVanL:
**What this PR does / why we need it**:
Uses wing and puppet to bootstrap vault/consul instances. Tarmak will now wait for convergence through wing for vault instances.

fixes #93 

```release-note
Vault instances are now bootstrapped and converged through wing + puppet
```

PR #664, opened 2018-12-19
SSH known hosts file management 

JoshVanL:
As per the proposal outlined here #643, this PR implements management of a clusters known hosts file.

During boot, wing will send it's public keys to a Lambda function which is verified and applied as tags to that instance. 
Tarmak will then use these tags as a source of truth and will self maintain the known hosts file.
The SSH client is now strictly enforcing the contents of this file when connecting to instances.

The `--wing-dev-mode` needs to be used during applies with this PR.

```release-note
Self maintain and strictly enforce the SSH `known_hosts` file.
```

/hold
Waiting for #494


terraform/amazon/modules/tagging_control/templates/tagging_control_policy.json
```diff
@@ -0,0 +1,27 @@
...
+  "Statement": [
+    {
+      "Sid": "",
+      "Effect": "Allow",
+      "Action": [
+        "ec2:CreateTags",
+        "ec2:DescribeNetworkInterfaces",
+        "ec2:CreateNetworkInterface",
+        "ec2:DetachNetworkInterface",
+        "ec2:DeleteNetworkInterface"
```
simonswine: Why are the Create/Detach/Delete necessary?

terraform/amazon/templates/bastion_user_data.yaml.template
```diff
@@ -73,13 +78,15 @@ write_files:
         chmod 755 /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy ;\
       fi ;\
       /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy download --output /opt/wing-$${WING_VERSION}/wing --sha256sums https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt  --signature-armored https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt.asc https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/wing_$${WING_VERSION}_linux_amd64'
+ {{- end }}
     ExecStartPre=/bin/sh -c '\
       set -e ;\
+      usermod -a -G ssh_keys wing ;\
```
simonswine: Is that really necessary to read the public host key?

terraform/amazon/templates/instance_pools/instance_pool_iam.tf.template
```diff
@@ -11,6 +11,11 @@ resource "aws_iam_instance_profile" "{{.TFName}}" {
   name = "${data.template_file.stack_name.rendered}.{{.TFName}}"
   role = "${aws_iam_role.{{.TFName}}.name}"
 }
+
+resource "aws_iam_role_policy_attachment" "{{.TFName}}_tagging_control_lambda_invoke" {
+  role       = "${aws_iam_role.{{.TFName}}.name}"
+  policy_arn = "${var.tagging_control_policy_arn}"
+}
```
simonswine: This fails to destroy for me:

```
During destroy:

DEBU[0176] Error: Error applying plan:  app=tarmak module=terraform std=err
DEBU[0176]                                               app=tarmak module=terraform std=err
DEBU[0176] 1 error occurred:                             app=tarmak module=terraform std=err
DEBU[0176]      * module.vault.aws_iam_policy_attachment.vault_tagging_control_lambda_invoke (destroy): 1 error occurred:  app=tarmak module=terraform std=err
DEBU[0176]      * aws_iam_policy_attachment.vault_tagging_control_lambda_invoke: [WARN] Error removing user, role, or group list from IAM Policy Detach csire-cluster-tagging-control-lambda-invoke:  app=tarmak module=terraform std=err
DEBU[0176] – NoSuchEntity: Policy arn:aws:iam::228615251467:policy/csire-cluster.tagging_control_lambda_invoke was not found.  app=tarmak module=terraform std=err
DEBU[0176]      status code: 404, request id: e3d171b2-1356-11e9-9e0c-0de57971cb89  app=tarmak module=terraform std=err
```

terraform/amazon/templates/jenkins_user_data.yaml.template
```diff
@@ -104,12 +104,48 @@ write_files:
...
+    RestartSec=10
+    Environment=WING_CLOUD_PROVIDER=amazon
+{{- if .WingDevMode }}
+    Environment=WING_VERSION="${wing_version}"
+    ExecStartPre=/bin/sh -c 'aws s3 cp "s3://${wing_binary_path}" /opt/wing-$${WING_VERSION}/wing; chmod 0755 /opt/wing-$${WING_VERSION}/wing'
+{{- else }}
+    Environment=AIRWORTHY_VERSION=0.2.0
+    Environment=AIRWORTHY_HASH=2d69cfe0b92f86481805c28d0b8ae47a8ffa6bb2373217e7c5215d61fc9efa1d
+    Environment=WING_VERSION=0.5.3
+    ExecStartPre=/bin/sh -c '\
```
simonswine: This should be a wing-install service that both units depend on

terraform/amazon/templates/jenkins_user_data.yaml.template
```diff
@@ -104,12 +104,48 @@ write_files:
...
+      if [ ! -x /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy ]; then \
+        mkdir -p /opt/airworthy-$${AIRWORTHY_VERSION} ;\
+        curl -sLo /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy https://github.com/jetstack/airworthy/releases/download/$${AIRWORTHY_VERSION}/airworthy_$${AIRWORTHY_VERSION}_linux_amd64 ;\
+        echo "$${AIRWORTHY_HASH}  /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy" | sha256sum -c ;\
+        chmod 755 /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy ;\
+      fi ;\
+      /opt/airworthy-$${AIRWORTHY_VERSION}/airworthy download --output /opt/wing-$${WING_VERSION}/wing --sha256sums https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt  --signature-armored https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/tarmak_$${WING_VERSION}_checksums.txt.asc https://github.com/jetstack/tarmak/releases/download/$${WING_VERSION}/wing_$${WING_VERSION}_linux_amd64'
+ {{- end }}
+    ExecStart=/bin/sh -c 'exec /opt/wing-$${WING_VERSION}/wing tag --environment "${tarmak_environment}"'
+    Type=notify
```
simonswine: In theory it should just exit after it has finished, Type=notify seems wrong, should be more like Type=oneshot

terraform/amazon/templates/vault_instances.tf.template
```diff
@@ -38,7 +38,7 @@ data "template_file" "vault" {
     puppet_tar_gz_bucket_dir = "${var.secrets_bucket}/${data.template_file.stack_name.rendered}/puppet-manifests"
 
     # These are only used in the template when running in Wing dev mode
-    wing_binary_path = "${var.secrets_bucket}/${var.wing_binary_path}"
+    wing_binary_path = "${var.secrets_bucket}/${data.template_file.stack_name.rendered}/${var.wing_binary_path}"
```
simonswine: Is that for different wing versions in a multi cluster environment?

terraform/amazon/templates/vault_instances.tf.template
```diff
@@ -99,6 +99,7 @@ resource "aws_instance" "vault" {
 
   lifecycle {
     ignore_changes = ["volume_tags"]
+    ignore_changes = ["tags"]
```
simonswine: What happens if we change the ebs volume tags?

terraform/amazon/modules/bastion/bastion_iam.tf
```diff
@@ -17,3 +17,28 @@ resource "aws_iam_policy_attachment" "bastion_additional_policy" {
   count      = "${length(var.bastion_iam_additional_policy_arns)}"
   policy_arn = "${element(var.bastion_iam_additional_policy_arns, count.index)}"
 }
+
+data "template_file" "wing_binary_read" {
```
simonswine: Can't we ruse the same for all instance types?

terraform/amazon/modules/bastion/bastion.tf
```diff
@@ -38,6 +42,10 @@ resource "aws_instance" "bastion" {
     volume_size = "${var.bastion_root_size}"
   }
 
+  lifecycle {
+    ignore_changes = ["tags"]
```
simonswine: This might have implications on other tags we use

terraform/amazon/templates/jenkins_user_data.yaml.template
simonswine: Ah sorry it is the jenkins instance :grimacing:

pkg/tarmak/assets/assets.go
```diff
@@ -8,4 +8,5 @@
 
 package assets
 
-//go:generate go-bindata -prefix ../../../ -pkg $GOPACKAGE -o assets_bindata.go ../../../terraform/amazon/modules/... ../../../terraform/amazon/templates/... ../../../puppet/... ../../../packer/...
+//go:generate go run ../../../cmd/tagging_control/main.go zip ../../../tagging_control.zip ../../../tagging_control_linux_amd64
```
simonswine: This is really ugly, not too sure if it break during the release process

terraform/amazon/modules/tagging_control/templates/tagging_control_policy.json
JoshVanL: These are required for lambda to access resources inside a VPC, although Detach should be removed.
https://docs.aws.amazon.com/lambda/latest/dg/vpc.html

```
The Lambda function execution role must have permissions to create, describe and delete ENIs. AWS Lambda provides a permissions policy, AWSLambdaVPCAccessExecutionRole, with permissions for the necessary EC2 actions (ec2:CreateNetworkInterface, ec2:DescribeNetworkInterfaces, and ec2:DeleteNetworkInterface) that you can use when creating a role. You can review the policy in the IAM console. Do not delete this role immediately after your Lambda function execution. There is a delay between the time your Lambda function executes and ENI deletion. If you do delete the role immediately after function execution, you are responsible for deleting the ENIs.
```

terraform/amazon/templates/bastion_user_data.yaml.template
JoshVanL: This is so wing can read the ssh private keys to sign the document, to prove we have them

pkg/tarmak/assets/assets.go
JoshVanL: Why would this break?

simonswine:
/unassign
/assign @JoshVanL 

Can you look at the rebase?

simonswine:
I think the release tarmak version might have some issue with this change. And obviously it complicates and slowes down the build process.

Other than that I seem to have a few problems where I need to delete the known_hosts file while terraform-tarmak is failing to connect to the bastion.

I am gonna merge it for now but this will be an area of stabilisation/testing

/approve
/lgtm

PR #682, opened 2019-01-15
Release public packer base images

simonswine:
```release-note
Release public AWS AMI images with packer
```


simonswine:
This was a quite tricky thing to achieve.

I had to manually create another base image to start those builds from. As the centos image is coming through the Marketplace and we cannot publish it ourselves

simonswine:
Think that is ready to go now

PR #705, opened 2019-01-25
Separate binaries and assets

simonswine:
Separate binaries and assets

* assets: packer, puppet and terraform code
* binaries: vendored tagging_control and wing linux amd64 binaries

Removing zip functionality from `tagging_control`

```release-note
NONE
```


JoshVanL:
/lgtm
/unassign
needs rebase
/assign @simonswine

PR #707, opened 2019-01-29
Remove VPC endpoints from being created in existing VPC

### LINKED ISSUE #706: VPC endpoint is created in an existing-vpc deployment

/kind bug

**What happened**:
An S3 vpc endpoint is created when deploying tarmak into an existing VPC.
This causes errors if the VPC endpoint exists as well as undefined behavior when that endpoint is configured.

JoshVanL:
fixes #706 

This removes this VPC endpoint from being created, along with a warning in the docs that it is strongly recommended that one exists. 

```release-note
S3 VPC endpoint is no longer created in deployments into an existing VPC
```

### DIFF
--- a/terraform/amazon/modules/network-existing-vpc/route53.tf
+++ b/terraform/amazon/modules/network-existing-vpc/route53.tf
@@ -5,7 +5,7 @@ resource "aws_route53_zone" "private" {
   comment       = "Hosted zone for private kubernetes in ${var.environment}"
 
   vpc {
-    vpc_id = "${aws_vpc.main.0.id}"
+    vpc_id = "${var.vpc_id}"
   }
 
   tags {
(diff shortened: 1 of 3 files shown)


terraform/amazon/modules/network-existing-vpc/route53.tf
```diff
@@ -5,7 +5,7 @@ resource "aws_route53_zone" "private" {
   comment       = "Hosted zone for private kubernetes in ${var.environment}"
 
   vpc {
-    vpc_id = "${aws_vpc.main.0.id}"
```
simonswine: Why did we not see this error?

JoshVanL: I guess we didn't use an existing VPC since we updated terraform and so used the new variable assignments. Something to look out for

PR #715, opened 2019-01-31
Use core-dns over kube-dns for clusters >= 1.10 

### LINKED ISSUE #422: Migration kube-dns => core-dns
We should migrate all cluster 1.10 and newer from kube-dns to core-dns.

* [ ] Scale down/remove kube-dns deployment
* [ ] Add Core-DNS deployment
* [ ] Rename service (?)

/kind feature

JoshVanL:
What this PR does / why we need it:
Use core-dns instead of kube-dns by default for clusters >= 1.10

Ensures that kube-dns is replaced by core-dns when upgrading >= 1.10 and visa-versa.

fixes #422

```release-note
Default to core-dns over kube-dns for clusters >= 1.10
```

/assign


JoshVanL:
Tested Prometheus and was scraping core-dns
/unassign
/assign @simonswine

simonswine:
Upgrade works well for me, thanks this is way cleaner @JoshVanL 

/lgtm

PR #683, opened 2019-01-16
Feature flags for calico with CRDs and optionally typha

JoshVanL:
fixes #435 

```release-note
Add Kubernetes API server as optional backend, as well as enabling Typha for Calico
```
/assign

PR #716, opened 2019-02-01
Feature flags for calico with CRDs and optionally typha

### LINKED ISSUE #435: Feature flags for calico with CRDs and eventually typha
Calico feature flags for CRDs (etcdless) deployment and if that is enabled there should be the option for typha. Default should remain the etcd deployment. The option is cluster wide in the `tarmak.yaml`

simonswine:
fixes #435 

```release-note
Add Kubernetes API server as optional backend, as well as enabling Typha for Calico
```
/assign


simonswine:
This is follow-up #683

simonswine:
I think we should also think about what we gonna do with the overlay etcd server, I think we should at least not set it up if we have kubernetes backend

PR #720, opened 2019-02-06
Fix heapster vertical pod autoscaler race condition

simonswine:
This was race conditions since the introduction of kube-addon-manager

```release-notes
NONE
```


JoshVanL:
Works for me
/lgtm

PR #735, opened 2019-02-11
Use role_policy_attachment for additional policies

JoshVanL:
Use role_policy_attachments for additional policies as per https://www.terraform.io/docs/providers/aws/r/iam_policy_attachment.html

```release-note
NONE
```
/assign @simonswine


simonswine:
Thanks that fixes it

/approve
/lgtm

PR #739, opened 2019-02-12
Fallback to pre-built AMI images if none privately built

### LINKED ISSUE #415: Provide prebuilt AMI image per region

JoshVanL:
If EBS is encrypted and no images have been privately built then fallback to using Jetstack's publicly built image for that version. 

fixes #415 

```release-note
Fallback to using Jetstack's pre-built public AMI if none are privately built.
```

/assign @simonswine

### DIFF
--- a/pkg/packer/packer.go
+++ b/pkg/packer/packer.go
@@ -10,6 +10,7 @@ import (
 
 	tarmakv1alpha1 "github.com/jetstack/tarmak/pkg/apis/tarmak/v1alpha1"
 	"github.com/jetstack/tarmak/pkg/tarmak/interfaces"
+	"github.com/jetstack/tarmak/pkg/version"
 )
 
 type Packer struct {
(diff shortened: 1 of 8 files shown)


pkg/packer/packer.go
```diff
@@ -116,5 +117,21 @@ func (p *Packer) IDs(encrypted bool) (map[string]string, error) {
...
+
+		image, err := p.tarmak.Cluster().Environment().Provider().DefaultImage(version)
+		if err != nil {
+			return nil, err
+		}
+
+		imagesChangeTime[image.BaseImage] = image.CreationTimestamp.Time
+		imageIDByName[image.BaseImage] = image.Name
+
+		p.log.Warnf("EBS is unencrypted so using Jestack's pre-built default image from version %s",
```
simonswine: Spelling of jetstack

JoshVanL: Good catch, not the best thing to misspell...

PR #757, opened 2019-02-25
Increases inotify limits on Kubernetes instances

JoshVanL:
Increase inotify watch limits to Kubernetes instances.
max_user_instances = 8192,
max_user_watches = 524288,

```release-note
Increase inotify watch limits on Kubelet instances
```

/assign @simonswine

### DIFF
--- a/puppet/modules/kubernetes/manifests/install.pp
+++ b/puppet/modules/kubernetes/manifests/install.pp
@@ -1,8 +1,5 @@
 # download and install hyperkube
-class kubernetes::install (
-  Integer $max_user_instances = $::kubernetes::max_user_instances,
-  Integer $max_user_watches = $::kubernetes::max_user_watches,
-){
+class kubernetes::install {
   include kubernetes
 
   $hyperkube_path = "${::kubernetes::_dest_dir}/hyperkube"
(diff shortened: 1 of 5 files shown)


puppet/modules/kubernetes/manifests/install.pp
```diff
@@ -19,4 +22,18 @@
...
+    path        => ['/usr/bin/', '/bin', '/usr/sbin'],
+  }
+
+  file{"${::kubernetes::params::sysctl_dir}/fs.conf":
+    ensure  => file,
+    mode    => '0640',
+    owner   => 'root',
+    content => template('kubernetes/fs.conf.erb'),
+    notify  => Exec['sysctl-system'],
+  }
```
simonswine: Can we do that as part of `::kubelet` instead of `::install`. 

Kubelet seems to be the part causing the issues with it's cadvisor integration.

puppet/modules/kubernetes/manifests/kubelet.pp
```diff
@@ -180,6 +183,27 @@
     $_ca_file = $ca_file
   }
 
+  exec {'sysctl-system':
+    command     => 'sysctl --system',
+    refreshonly => true,
+    path        => ['/usr/bin/', '/bin', '/usr/sbin'],
```
simonswine: If you use `$::kubernetes::path` here it should also work on ubuntu as they have the `sysctl` bin in `/sbin`

JoshVanL: Thanks, missed this yesterday

PR #801, opened 2019-04-26
Increases vault max_lease_ttl to 5 years

JoshVanL:
Signed-off-by: JoshVanL <vleeuwenjoshua@gmail.com>

/assign @simonswine 

```release-note
Increase vault's max_lease_ttl to 5 years
```


simonswine:
Why is this necessary? Is it not enough to do that using `vault auth tune /token`?

JoshVanL:
@simonswine  Yes should be
/close

PR #802, opened 2019-04-26
Adds centos to Makefile for tests

JoshVanL:
```release-note
NONE
```

/assign @simonswine

### DIFF
--- a/puppet/modules/vault_server/Makefile
+++ b/puppet/modules/vault_server/Makefile
@@ -13,3 +13,6 @@ acceptance-centos: bundle_install
 
 acceptance-ubuntu: bundle_install
 	bundle exec rake beaker:ubuntu_1604_single_node
+
+acceptance-three-node: bundle_install
+	bundle exec rake beaker:3n-cluster
--- a/puppet/modules/vault_server/Makefile
+++ b/puppet/modules/vault_server/Makefile
@@ -6,11 +6,10 @@ verify: bundle_install
 bundle_install:
 	bundle install --path $(BUNDLE_DIR)
 
-acceptance: bundle_install
+acceptance: accetance-centos
+
+acceptance-centos: bundle_install
 	bundle exec rake beaker:default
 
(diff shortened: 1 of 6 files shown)


puppet/modules/vault_server/Makefile
```diff
@@ -6,11 +6,10 @@ verify: bundle_install
 bundle_install:
 	bundle install --path $(BUNDLE_DIR)
 
-acceptance: bundle_install
-	bundle exec rake beaker:default
+acceptance: accetance-centos
 
-# provide backwards compatiblity
-acceptance_three_node: acceptance-three-node
```
simonswine: I guess we never ran those?

/assign @JoshVanL 
/unassign

simonswine: Also doesn't make too much sense, as it's active/passive/passive

JoshVanL:
Unsure why this is failing, it's passing locally. Will investigate further

puppet/modules/vault_server/templates/vault.service.erb
```diff
@@ -5,6 +5,7 @@ After=consul.service vault-assets.service
 
 [Service]
 Environment=AWS_REGION=<%= @region %>
+ExecStartPre=/bin/sh -c "while true; do STATUS=$(curl -s -o /dev/null -k -w '%{http_code}' --request GET http://localhost:8500/v1/health/node/<%= @hostname %>); if [ $STATUS -eq 200 ]; then break; else sleep 5; fi; done"
```
simonswine:
```
// check for acl readiness
 req, err = http.NewRequest("GET", fmt.Sprintf("%s/v1/acl/token/self", c.HTTPHostAddr), nil)
 if err != nil {
        return fmt.Errorf("error building acl readiness request: %v", err)
}
```

Node healthy does not mean the ACL system is ready. Can you please test it like that?

Also the whole hostname thing is quite ugly

